PR #14, opened 2018-11-06
Missing $ for inline equation

mlxd:
The documentation for the ContinuousOracle type (https://docs.microsoft.com/en-us/qsharp/api/canon/microsoft.quantum.canon.continuousoracle?view=qsharp-preview) has a missing $ symbol to close the inline equation. This change is modified in the quantum-docs-pr as well as here


mlxd:
As the PR in the docs was closed, I will assume this is to be closed also.

PR #15, opened 2018-11-24
Math Functions

jean002:


martinquantum:
@jean002, thank you for your contribution. While the factorial function might be of interest as a quantum operation, i.e., a unitary operator that implements $\ket{x}\ket{y} \mapsto \ket{x}\ket{y \oplus x!}$, with suitably chosen registers for $\ket{x}$ and $\ket{y}$, the implementation of the classical factorial function does not seem like a useful libraries contribution. Therefore, unfortunately, we cannot accept this contribution.

martinquantum:
Closing this PR.

PR #19, opened 2018-12-19
Fix "# Example" tag in documentation comments

tcNickolas:
Standalone "# Example" tag is currently ignored by the parser; the currently supported syntax is "# Remarks / ## Example"


alan-geller:
If OPS will actually support the `example` tag, I can easily change the doc gen tool to generate that tag from `# Example`, and even `## Example` for now...

alan-geller:
Merging -- the build failure is a red herring, it was due to an external web site being down.

PR #28, opened 2019-02-09
Bugfix for RandomReal, added a unit test for it

jclapis:
Hi all,

I'm working with the QDK and noticed a bug with RandomReal. It was generating numbers between 0 and 2, but the documentation says that it should be between 0 and 1. I fixed it here and added a unit test for the RandomReal function as well since it didn't have one.


jclapis:
@cgranade @vadym-kl 
Hello folks, just checking back on this. Anything you need me to take care of before reviewing?

pensono:
Should the random number generator be seeded so the test is deterministic?

anpaz-msft:
@pensono, there is nothing really random about this test, no need to make the it deterministic.

anpaz-msft:
@jclapis, sorry for the delay. it looks good. merging now and should make it in time for our next release.

PR #32, opened 2019-02-22
Added new transformed operation fns.

cgranade:
This PR adds new functions and operations for transforming operation inputs using functions. Marked as draft until I finish adding tests.

### DIFF
--- a/Canon/src/Combinators/Transformed.qs
+++ b/Canon/src/Combinators/Transformed.qs
@@ -1,0 +1,311 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Canon {
+
+    // #region No functors
+
+    /// # Summary
+    /// Given an operation that accepts some input, a function that
+    /// returns an output compatible with that operation, and an input to that
+    /// function, applies the operation using the function to transform the
+    /// input to a form expected by the operation.
+    ///
+    /// # Input
+    /// ## fn
+    /// A function that transforms the given input into a form expected by the
+    /// operation.
+    /// ## op
+    /// The operation to be applied.
+    /// ## input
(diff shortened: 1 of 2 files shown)


Canon/src/Combinators/Transformed.qs
```diff
@@ -0,0 +1,311 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Canon {
```
bettinaheim: Should this be in Canon? Looks like something that should go into Core at some point (potentially being a backing function for an operator in the future).

bettinaheim: Also, I assume you'll make a pass to replace similar utils in the libraries to use this one instead in a separate PR?

cgranade: @bettinaheim: Yes, my intent is to use this to replace or to implement a lot of existing patterns throughout the libraries. In particular, a lot of the arithmetic BigEndian ↔ LittleEndian logic is a special case of this new functionality.

bettinaheim: Sounds good.

cgranade: With respect to moving into Core, I'd like to defer that to a future PR if that's alright?

cgranade:
@anpaz-msft, would you happen to know what's up with the unit test failures? I'm not sure why this is causing an exception in the simulation runtime. Thanks!

Canon/src/Combinators/Transformed.qs
bettinaheim: Sure, just wanted to point it out. :)

cgranade:
This has been obviated by #39, so going on and closing this PR.

Issue #31, opened 2019-02-21
Cross-platform Python host functionality


cgranade:
We should enhance the existing Python interoperability feature to target cross-platform support as well as Windows.

cpalmer2020:
We now support cross platform interop.  See the release notes for February.  
https://docs.microsoft.com/en-us/quantum/install-guide/python?view=qsharp-preview

PR #36, opened 2019-02-26
Cross-platform client for Python host

cgranade:
This PR completes the client side of the host support tracked in #31. The README needs a bit of cleanup once the corresponding documentation has been added to quantum-docs-pr, but this should otherwise be good to go.


Interoperability/python/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+On top of simulation, you can also do quantum resources estimation including 
+the count of primitive operations used by the algorithm and the number of required qubits.
+For this, invoke the `trace` method on the operation:
+```python
+r = HelloAgain.trace(count=5, name="Counting")
+```
+
+You may use the `qsharp.print_tracer_counts` method to print the trace results to the console:
```
anpaz-msft: This is not the case anymore, the results contain already a dictionary with the resources calculated.

Interoperability/python/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+
+You can now also compile Q# operations on the fly from Python
+and simulate them.
+To create an operation on the fly, the `qsharp` module exports a `compile` method which receives two parmaters
+* id: a unique identifier of this snippet.
+* source: a valid Q# code snippet with the operation definition, for example:
+```python
+hello = qsharp.compile("snippet_1", """
```
anpaz-msft: the id parameter is now gone.

Interoperability/python/src/Operations.qs
```diff
@@ -0,0 +1,36 @@
+﻿/// Q# code should be in one or more .qs files that live
```
anpaz-msft: this file should not go here, right?

Interoperability/python/src/driver.py
```diff
@@ -0,0 +1,89 @@
+#!/bin/env python
```
anpaz-msft: similar this file should not be here.

Interoperability/python/src/qsharp/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+    ops = [
+        QSharpCallable(op, "snippets")
+        for op in client.compile(code)
+    ]
+    if len(ops) == 1:
+        return ops[0]
+
+def reload() -> None:
+    """
+    Reloads the current IQ# workspace, discarding any current snippets and
```
anpaz-msft: I don't think we're discarding current snippets (yet).

anpaz-msft:
Because you removed `step-build-python.yml` you'll need to update `build.yml`.

Interoperability/python/src/Operations.qs
cgranade: Good catch, moving this and `driver.py` to the samples repo.

Interoperability/python/src/qsharp/__init__.py
cgranade: Good catch, I've fixed the docstring accordingly.

Interoperability/python/README.md
cgranade: Honestly, most of the README can be deferred to the docs now, I think. I'll take a pass to focus it on what makes the most sense here.

Interoperability/python/README.md
cgranade: Resolved by other changes to README. Thanks!

cgranade:
> Because you removed `step-build-python.yml` you'll need to update `build.yml`.

Ah, good point. I thought I'd done so, but must have missed it, sorry.

PR #42, opened 2019-03-15
Arithmetic

cryptosidh:
Added code for integer adders and greater than comparison with tests.


cgranade:
Wonderful, thank you so much for the feature, @cryptosidh!

PR #55, opened 2019-04-05
Add unitary coupled-cluster state preparation and update Broombridge

guanghaolow:
This is an update to the chemistry library focusing on Unitary Coupled-Cluster (UCC) methods. Once a UCC state is specified in Broombridge with a label, e.g. `"UCCSD |G>"`, it may be selected similar to how any other stately is current chosen through this label.

Main changes:
- Update Broombridge schema to allow specifying a UCC state
- Implemented Broombridge v0.1 to v0.2 updater to maintain backwards compatibility
- Add UCC parsing to chemistry library
- Add UCC quantum circuit 
- Add optional configuration file to specify low-level options. Currently, the only option is choosing indexing convention from a (orbital, spin) index to an integer.

Other changes:
- General cleanup.


Chemistry/src/DataModel/ConfigurationFile.cs
```diff
@@ -0,0 +1,65 @@
...
+        public Config()
+        {
+            indexConvention = IndexConvention.Default;
+            truncationThreshold = TruncationThreshold.Default;
+        }
+
+        /// <summary>
+        /// Choose indexing convention from spin-orbital index to an integer.
+        /// </summary>
+        public IndexConvention.Type indexConvention;
```
cgranade: In general, public properties should start with initial capital letters.
```suggestion
        public IndexConvention.Type IndexConvention;
```

Chemistry/src/DataModel/ConfigurationFile.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Available indexing convention from spin-orbital index to an integer.
+        /// </summary>
+        public class IndexConvention : SpinOrbital.Config.IndexConvention { }
+
+        /// <summary>
+        /// Threshold hold below which to truncate Hamiltonian coefficients.
```
cgranade:
```suggestion
        /// Threshold below which to truncate Hamiltonian coefficients.
```

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+    {
+        /// <summary>
+        /// Class containing Broombridge version number management.
+        /// </summary>
+        public static class Version
+        {
+            /// <summary>
+            /// Enumerable item for Broombridge version numbers.
+            /// </summary>
+            public enum Number
```
cgranade: I'd suggest against nesting classes and enums too deeply.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+            {
+                v0_1 = 0, v0_2 = 1
+            }
+
+            /// <summary>
+            /// Dictionary from version number strings to version number types.
+            /// </summary>
+            public static Dictionary<string, Number> VersionNumberDict = new Dictionary<string, Number>()
```
cgranade: It may be worth making this protected or internal, as I'm not sure that this is really useful to have as part of the public API of the `Version` class. The public interface is largely defined by the `ParseVersionNumber` method below, I think.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+                {"0.1", Number.v0_1 },
+                {"0.2", Number.v0_2 }
+            };
+
+            /// <summary>
+            /// Parse version number string
+            /// </summary>
+            /// <param name="versionNumber">Version number string</param>
+            /// <returns>Version number in enum `Number` format.</returns>
+            public static Number ParseVersionNumber(string versionNumber)
```
cgranade: An idiomatic C# pattern would be to have a `TryParse...` method with an `out` parameter to handle the case in which `versionNumber` isn't a valid version number. Under that idiom, this would likely be `public static bool ParseVersionNumber(string versionNumber, out Number parsedVersion)`.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+            }
+        }
+
+        // Implement backwards compatibility by converting v0.1 to v0.2.
+        // Only parse v0.2.
+
+        /// <summary>
+        /// Broombridge deserializers
+        /// </summary>
+        public static class Deserialize
```
cgranade: Generally, classes are named as nouns to represent that classes are kinds of things. For instance, I'd suggest `Deserializer` here instead.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+            /// </summary>
+            /// <param name="filename">Broombridge file address.</param>
+            /// <returns>Version number of Broombridge file</returns>
+            public static Version.Number GetVersionNumber(string filename)
+            {
+                using (var reader = File.OpenText(filename))
+                {
+                    var deserializer = new DeserializerBuilder().Build();
+                    var data = deserializer.Deserialize<Dictionary<string, object>>(reader);
+                    var format = (Dictionary<object, object>)data["format"];
```
cgranade: It may also be useful to look at the value of the `$schema` key, as the presence of that key is guaranteed by the JSON Schema specification. By contrast, we may not have a `format` key in the future.

cgranade: The casts here can fail in surprising ways, it may be better to use an `as` operator together with null coallescing operators.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+                    var version = (string)format["version"];
+                    return Version.ParseVersionNumber(version);
+                }
+            }
+
+            /// <summary>
+            /// Returns deserializer Broombridge data strauture.
```
cgranade: I'm not sure I understand this comment, sorry?

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+                {
+                    return v0_2(filename);
+                }
+                else {
+                    throw new System.InvalidOperationException("Unrecognized Broombridge version number.");
+                }
+            }
+
+            /// <summary>
+            /// Broombridge v0.1 deserializer
```
cgranade: This summary doesn't really explain what this method does?

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+            /// </summary>
+            /// <param name="filename">Broombridge filename to deserialize</param>
+            /// <returns>Deserialized Broombridge v0.1 data.</returns>
+            public static V0_1.Data v0_1(string filename)
```
cgranade: I'd suggest a more informative name here, perhaps.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+                    var deserializer = new DeserializerBuilder().Build();
+                    return deserializer.Deserialize<V0_2.Data>(reader);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Broombridge serializers
+        /// </summary>
+        public static class Serialize
```
cgranade: Similarly, I'd suggest Serializer here in preference to Serialize.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+
+
+        public static class Update
+        {
+            /// <summary>
+            /// Converts v0.1 Broombridge to v0.2.
+            /// </summary>
+            /// <param name="input">Source Broombridge in v0.1 format.</param>
+            /// <returns>Converted Broombridge in v0.2 format.</returns>
+            public static V0_2.Data Data(V0_1.Data input)
```
cgranade: It seems like a lot of these methods are public, but aren't really designed to be called by users directly?

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+                    }
+
+                    output.ProblemDescription.Add(problemDescription);
+                }
+
+                return output;
+            }
+        }
+
+        public static class DataStructures
```
cgranade: It may be reasonable to break this file into separate files.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+            }
+        }
+
+        #endregion
+    }
+    // Parts of this might be merged intro Broombridge parsing due to overlap.
+
+
+    /// <summary>
+    /// Broombridge with type information parsed, and only relevant metadata kept.
```
cgranade: I'm not sure I understand this summary, sorry?

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+                state.Superposition = clusterTerms.ToArray() ;
+            }
+            else
+            {
+                throw new System.ArgumentException($"initial state `{state.Label}` is not recognized or implemented.");
+            }
+            return state;
+        }
+
+        internal static FermionHamiltonian.StateType ParseInitialStateMethod(string state)
```
cgranade: Using a static readonly immutable dictionary may be easier than nested `if`/`else if` blocks with redundant calls to `ToLowerInvariant`.
- https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2.withcomparers?view=netcore-2.2#System_Collections_Immutable_ImmutableDictionary_2_WithComparers_System_Collections_Generic_IEqualityComparer__0__
- https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparer.invariantcultureignorecase?view=netcore-2.2

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+        }
+
+        public static ((Double, Double), FermionTerm) ParseInputState(List<string> superpositionElement, SpinOrbital.Config.IndexConvention.Type indexConvention)
+        {
+            var amplitude = Double.Parse(superpositionElement.First(), System.Globalization.CultureInfo.InvariantCulture);
+            var initialState = superpositionElement.Last();
+            var ca = new List<Int64>();
+            var so = new List<SpinOrbital>();
+
+            for (int i = 1; i < superpositionElement.Count() - 1; i++)
```
cgranade: Since the index isn't used in the loop, I'd suggest a `foreach` over `superpositionElement`.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
```diff
@@ -0,0 +1,808 @@
...
+            else
+            {
+                throw new System.ArgumentException($"{input} is not valid Polish notation");
+            }
+        }
+    }
+    
+
+
+    public static partial class Extensions
```
cgranade: These should definitely be split into their own file, and should be referenced if possible rather than duplicated.

Chemistry/src/DataModel/FermionHamiltonianStatePrep.cs
```diff
@@ -27,21 +27,30 @@ public partial class FermionHamiltonian
...
         /// </summary>
-        public List<InputState> InputStates = new List<InputState>();
+        public Dictionary<string, InputState> InputStates = new Dictionary<string, InputState>();
+
+        /// <summary>
+        /// Enum over valid input state types.
+        /// </summary>
+        public enum StateType
+        {
+            Default = 0, Single_Configurational = 1, Sparse_Multi_Configurational = 2, Unitary_Coupled_Cluster = 3
```
cgranade:
```suggestion
            Default = 0, SingleConfigurational = 1, SparseMultiConfigurational = 2, UnitaryCoupledCluster = 3
```

Chemistry/src/DataModel/OrbitalIntegral.cs
```diff
@@ -207,22 +207,37 @@ public SpinOrbital[][] EnumerateSpinOrbitals()
...
+                    var spin = (Spin)((idx / nSpins.Pow(idxOrbital)) % nSpins);
+                    spinOrbitalArray[fst] = new SpinOrbital(OrbitalIndices[fst], spin, indexConvention);
+                    spinOrbitalArray[lst] = new SpinOrbital(OrbitalIndices[lst], spin, indexConvention);
                 }
                 spinOrbitalArrayOfArray[idx] = spinOrbitalArray;
             }
             return spinOrbitalArrayOfArray;
         }
+
+        public bool Equals(OrbitalIntegral x)
```
cgranade:
```suggestion
        public override bool Equals(object x)
```

You'll likely want something like `if (x is OrbitalIntegral integral) { /* do something with integral here */ } else { return false; }` as well.

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+
+namespace Microsoft.Quantum.Chemistry.JordanWigner {
+    open Microsoft.Quantum.Simulation;
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Extensions.Math;
+    open Microsoft.Quantum.Chemistry;
+    open Microsoft.Quantum.Arrays;
+
+	/// # Summary
```
cgranade: Tabs vs spaces.

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+    ///
+    /// # Input
+    /// ## nFermions
+    /// Number of fermions in the system.
+	/// ## idxFermions
+	/// fermionic operator indices.
+    ///
+    /// # Output
+    /// Bitstring `Bool[]` that is `true` where a `PauliZ` should be applied.
+	function _ComputeJordanWignerBitString_(nFermions: Int,  idxFermions: Int[]) : Bool[] {
```
cgranade: I think the underscores on both sides of the name are likely redundant. Just need a leading underscore to mark this function as private.

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+    open Microsoft.Quantum.Arrays;
+
+	/// # Summary
+    /// Computes Z component of Jordan-Wigner string between
+	/// fermion indices in a fermionic operator with an even
+	/// number of creation / annihilation operators.
+    ///
+    /// # Input
+    /// ## nFermions
+    /// Number of fermions in the system.
```
cgranade:
```suggestion
    /// The number of fermions in the system.
```

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+	/// ## idxFermions
+	/// fermionic operator indices.
```
cgranade:
```suggestion
	/// Indices of the fermionic operator.
```

An `# Example` section describing how this input is understood may also be helpful; for a private function, may not be needed, though.

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+    ///
+    /// # Output
+    /// Bitstring `Bool[]` that is `true` where a `PauliZ` should be applied.
```
cgranade:
```suggestion
    /// A bit string represented as an array `zComponent` where `zComponent[idx]` is `true` if and only if a `Z` operation should be applied at the index `idx`.
```

An example may similarly be helpful here.

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+			fail $"ComputeJordanWignerString failed. `idxFermions` must contain an even number of terms.";
+		}
+		
+		mutable zString = new Bool[nFermions];
+		for(fermionIdx in idxFermions){
+			if(fermionIdx >= nFermions){
+				fail $"ComputeJordanWignerString failed. fermionIdx {fermionIdx} out of range.";
+			}
+			for(idx in 0..fermionIdx){
+				set zString[idx] = zString[idx] ? false | true;
```
cgranade:
```suggestion
				set zString[idx] = not zString[idx];
```
If I understand right, the purpose of this loop is to flip all bits at or below `fermionIdx` for each `fermionIdx` in the input array?

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+    ///
+    /// # Output
+    /// Bitstring `Bool[]` that is `true` where a `PauliZ` should be applied.
+	function _ComputeJordanWignerBitString_(nFermions: Int,  idxFermions: Int[]) : Bool[] {
+		if(Length(idxFermions) % 2 != 0){
+			fail $"ComputeJordanWignerString failed. `idxFermions` must contain an even number of terms.";
+		}
+		
+		mutable zString = new Bool[nFermions];
+		for(fermionIdx in idxFermions){
```
cgranade:
```suggestion
		for (fermionIdx in idxFermions) {
```

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+			if(fermionIdx >= nFermions){
+				fail $"ComputeJordanWignerString failed. fermionIdx {fermionIdx} out of range.";
+			}
+			for(idx in 0..fermionIdx){
+				set zString[idx] = zString[idx] ? false | true;
+			}
+		}
+		
+		for(fermionIdx in idxFermions){
```
cgranade:
```suggestion
		for (fermionIdx in idxFermions) {
```

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -0,0 +1,291 @@
...
+			set zString[fermionIdx] = false;
+		}
+		return zString;
+	}
+
+	// Identical to `_ComputeJordanWignerBitString_`, except with the map  
+	// false -> PauliI and true -> PauliZ
+	function _ComputeJordanWignerPauliZString_(nFermions: Int,  idxFermions: Int[]) : Pauli[] {
```
cgranade: I think this can be replaced with a call to `BoolArrayAsPauli` (which should be renamed to `BoolArrayAsPauliArray`, that's a bug, sorry — will fix it).

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Move up one level

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Fixed

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: That is correct

Chemistry/src/DataModel/ConfigurationFile.cs
guanghaolow: Noted

Chemistry/src/DataModel/ConfigurationFile.cs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Fixed, but `Serializer` is already used by the serialzier library. Renamed instead to Dereializers and Serializers.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Renamed to`DeserializeBroombridgev0_1`

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Serializer is already taken. Renamed to Serializers

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Fixed in chemistry library

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Needed to set it as public for unit tests

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Broken up

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
cgranade: Generally, the preferred approach is to then mark it as `internal` and use the `InternalsVisibleTo` attribute.

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Added example

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Elaborated to `Deserialize Broombridge v0.1 from a file into the Broombridge v0.1 data structure.`

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Split

Chemistry/src/DataModel/OrbitalIntegral.cs
guanghaolow: Fixed

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Fixed here and elsewhere

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Noted

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Fixed

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
guanghaolow: Added example

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Fixed by replacing this with checking the $schema key, as suggested in [above](https://github.com/Microsoft/QuantumLibraries/pull/55#discussion_r272774487)

Chemistry/src/DataModel/ConfigurationFile.cs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Using $schema key now

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Deleted this class

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Leaving this alone for now as Andres set many of the data structure access properties to `public` for python interop.

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Fixed

Chemistry/src/DataModel/DataFormats/Broombridge.cs
guanghaolow: Now using the index in the lop

guanghaolow:
Thanks for the great comments Chris! I'll be addressing most of them in branch gulow/UsabilityImprovements. I'll make a pull request of that onto this shortly.

cpalmer2020:
@guanghaolow Shouldn't this be on the 0.7 branch?

guanghaolow:
This is a version of UCCSD that makes minimal changes to v0.6. Closing this PR as these changes have been incorporated into the v0.7 chemistry library.

Issue #59, opened 2019-04-08
qsharp.compile doesn't return the list of operations if more than one is provided


anpaz-msft:
When more than one operation is included in the snippet to compile, compile is not returning any operation, for example:

```python
import qsharp;

ops = qsharp.compile("""
    operation PrepareBell(AliceQ : Qubit, BobQ : Qubit) : Unit {
        H(AliceQ);
        CNOT(AliceQ, BobQ);
    }
    operation ExtractMessage(AliceQ : Qubit, BobQ : Qubit) : Unit {
        let resultAlice = M(AliceQ);
        let resultBob = M(BobQ);
        Message($"Result: {resultBob}, {resultAlice}");
    }
""")

print(ops)
```

Should print
```
[<Q# callable ExtractMessage>, <Q# callable PrepareBell>]
```

but it's currently empty.

anpaz-msft:
Fixed on master. Needs to be released yet.

PR #60, opened 2019-04-08
Missing return statement when compiling more than one Q# snippet from Python.

anpaz-msft:
Fixing #59


anpaz-msft:
This fixes #59

PR #63, opened 2019-04-10
Adding Jupyter Magic for chemistry v2

anpaz-msft:
Cleans up the chemistry magic, adds tests.


Chemistry/src/DataModel/Magic/BroombridgeMagic.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.Jupyter.Core;
+using Microsoft.Quantum.Chemistry.Broombridge;
+
```
guanghaolow: We should probably create a new project for parts that can invoke the library as a black-box, e.g. Magic, Workflows, Scripts.

anpaz-msft: I thought about it, except we would either need to also create a new nuget package, or move the nuget package definition to that project.

Chemistry/src/DataModel/Magic/FermionHamiltonianMagic.cs
```diff
@@ -0,0 +1,173 @@
...
+            /// </summary>
+            public List<(HermitianFermionTerm, double)> fermionTerms { get; set; }
+        }
+
+        /// <summary>
+        /// Simply calls AddRange on the hamiltonian to add each term from the list of fermionTerms
+        /// </summary>
+        public ExecutionResult Run(string input, IChannel channel)
+        {
+            var args = JsonConvert.DeserializeObject<Arguments>(input);
```
cgranade: We should check if `input` is null or empty before trying to deserialize it.

Chemistry/src/DataModel/Magic/InputStateMagic.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    /// <summary>
+    /// Loads Broombridge electronic structure problem and returns fermion Hamiltonian.
+    /// </summary>
+    public class InputStateMagic : MagicSymbol
+    {
+        public InputStateMagic()
+        {
+            this.Name = $"%inputstate-load";
+            this.Documentation = new Documentation() { Summary = "Loads Broombridge electronic structure problem and returns selected input state." };
```
cgranade:
```suggestion
            this.Documentation = new Documentation() { Summary = "Loads an electronic structure problem from a Broombridge file or a JSON-serialized problem description, and returns the selected input state." };
```

Chemistry/src/DataModel/Magic/InputStateMagic.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        /// <summary>
+        /// List of arguments
+        /// </summary>
+        public class Arguments
+        {
+            /// <summary>
+            /// The name of a .yaml file with a broombridge schema.
+            /// </summary>
+            public string fileName { get; set; }
```
cgranade: Should these be upper-case as well?

Chemistry/src/DataModel/Magic/InputStateMagic.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// The ProblemDescription can come from a .yaml broombridge file or can be passed down as parameter.
+        /// If the wavefunctionLabel is specified, it will return the corresponding inputstate from the
+        /// ProblemDescription; if the wavefunctionLabel is not specified, then it returns 
+        /// the Hartree--Fock state.
+        /// </summary>
+        public ExecutionResult Run(string input, IChannel channel)
+        {
+            if (string.IsNullOrWhiteSpace(input))
+            {
+                channel.Stderr("Please provide the name of a broombridge file or a problemDescription to load the FermionHamiltonian from.");
```
cgranade:
```suggestion
                channel.Stderr("Please provide the name of a Broombridge file or a problem description to load the fermion Hamiltonian from.");
```

Since there's a lot of overlap in error messages, maybe we should consider moving those out to res strings?

Chemistry/src/DataModel/Magic/InputStateMagic.cs
```diff
@@ -0,0 +1,109 @@
...
+                ? CreateHartreeFockState(problemData, args.indexConvention)
+                : problemData.ToWavefunctions(args.indexConvention)[args.wavefunctionLabel];
+
+            return inputState.ToExecutionResult();
+        }
+
+        /// <summary>
+        /// Creates a HartreeFork state from the FermionHamiltonian associated with the given ProblemDescription
+        /// </summary>
+        public static InputState CreateHartreeFockState(CurrentVersion.ProblemDescription problemData, SpinOrbital.IndexConvention indexConvention)
```
cgranade: This method doesn't reference anything about the magic command or its arguments. Given that, should it be in magic, or should it be a method / extension method on ProblemDescription?

Chemistry/src/DataModel/Magic/InputStateMagic.cs
```diff
@@ -0,0 +1,109 @@
...
+            return fermionHamiltonian.GreedyStatePreparation(problemData.NElectrons);
+        }
+
+
+        /// <summary>
+        /// Selects the ProblemDescription from the given arguments.
+        /// If the fileName is specified, it will try to load the Broombridge data from the file
+        /// and will use the first ProblemDescription, otherwise, the problemDescription in the arguments is used.
+        /// </summary>
+        protected virtual CurrentVersion.ProblemDescription SelectProblemDescription(Arguments args)
```
cgranade: This doesn't reference an instance of the magic command, so should it be part of arguments or static?

Chemistry/src/DataModel/Magic/InputStateMagic.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
```
cgranade: This can be similarly simplified, or even combined with the above method.

Chemistry/tests/DataModelTests/MagicTests/ChemistryEncodeMagicTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        public (ChemistryEncodeMagic, MockChannel) Init() =>
+            (new ChemistryEncodeMagic(), new MockChannel());
+
+        [Fact]
+        public void EncodeBroombridge()
+        {
+            var (magic, channel) = Init();
+            var fileName = "Broombridge/broombridge_v0.2.yaml";
+
+            Assert.Equal("%chemistry-encode", magic.Name);
```
cgranade: None of the other magic commands are prefaced with `chemistry`; we should be consistent on that.

Chemistry/tests/DataModelTests/MagicTests/MockChannel.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+
+using Microsoft.Jupyter.Core;
+
+namespace Microsoft.Quantum.Chemistry.Tests
+{
+    public class MockChannel : IChannel
```
cgranade: This is generically useful enough it might should be a jupyter-core feature.

Interoperability/python/src/qsharp/tests/Operations.qs
```diff
@@ -0,0 +1,43 @@
...
+/// Q# code should be in one or more .qs files that live 
+/// in the same directory as the python clasical driver.
+///
+
+namespace Microsoft.Quantum.SanityTests
+{
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+
+    /// # Summary:
```
cgranade:
```suggestion
    /// # Summary
```

Interoperability/python/src/qsharp/tests/Operations.qs
```diff
@@ -0,0 +1,43 @@
...
+/// in the same directory as the python clasical driver.
+///
+
+namespace Microsoft.Quantum.SanityTests
+{
+    open Microsoft.Quantum.Primitive;
+    open Microsoft.Quantum.Canon;
+
+    /// # Summary: 
+    ///     The simplest program. Just generate a debug Message on the console.
```
cgranade:
```suggestion
    /// The simplest program. Just generate a debug Message on the console.
```

Chemistry/src/DataModel/Magic/FermionHamiltonianMagic.cs
```diff
@@ -0,0 +1,173 @@
...
+            FermionHamiltonian fermionHamiltonian = new FermionHamiltonian();
+
+            return fermionHamiltonian.ToExecutionResult();
+        }
+    }
+
+    /// <summary>
+    /// Adds terms to a fermion Hamiltonian.
+    /// </summary>
+    public class FermionHamiltonianAddTermsMagic : MagicSymbol
```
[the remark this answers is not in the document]
guanghaolow: The ideal solution would be to write a python-side function to modify the datastructures living on python there. Is there a good way to automatically covert or compile the relevant parts of the C# code to python? It would not even need to be human readable.

Chemistry/src/DataModel/Magic/BroombridgeMagic.cs
anpaz-msft: Turns out we actually need it, otherwise we have a chicken and egg problem with IQ# build. I have extract the magic into its own solution/package

Chemistry/src/DataModel/Magic/ChemistryEncodeMagic.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Microsoft.Quantum.Chemistry.Magic
+{
+    /// <summary>
+    /// Encodes a fermion Hamiltonian and wavefunction ansatz into a format consumable by Q#.
+    /// </summary>
+    public class ChemistryEncodeMagic : MagicSymbol
+    {
+        public ChemistryEncodeMagic()
+        {
+            this.Name = $"%chemistry-encode";
```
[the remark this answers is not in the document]
anpaz-msft: they are not very useful, indeed.
However, I think it's fine. This is only available for someone using  the chemistry library. It's unclear to me how will they even be discovered.

Chemistry/src/DataModel/Magic/FermionHamiltonianMagic.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        /// <summary>
+        /// Adds terms to a fermion Hamiltonian.
+        /// </summary>
+        public FermionHamiltonianAddTermsMagic()
+        {
+            this.Name = $"%fh-add_terms";
```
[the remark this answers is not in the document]
anpaz-msft: I'm using underscore for separating works, and using dashes to seperate commands, so in this case, this is the "add terms" command from the "fh" (FermionHamiltonian) magic.

Chemistry/src/DataModel/Magic/InputStateMagic.cs
anpaz-msft: Because these are only used for json encoding I'm skipping using capitals.

Chemistry/src/DataModel/Magic/InputStateMagic.cs
anpaz-msft: Yes, we should. :)

Chemistry/src/DataModel/Magic/InputStateMagic.cs
anpaz-msft: I'm leaving it as a virtual protected method in case someone wants to create their own magic that overrides the default selection of the problem.

Chemistry/src/DataModel/Magic/InputStateMagic.cs
anpaz-msft: Created an extension method for ProblemDescription.

Chemistry/tests/DataModelTests/MagicTests/ChemistryEncodeMagicTests.cs
anpaz-msft: that's because this one is not specific to any data structure.

anpaz-msft: and i didn't want to name it just "%encode"

Chemistry/src/DataModel/Magic/FermionHamiltonianMagic.cs
```diff
@@ -0,0 +1,173 @@
...
+            
+            return fermionHamiltonian.ToExecutionResult();
+        }
+
+        /// <summary>
+        /// Selects the ProblemDescription from the given arguments.
+        /// If the fileName is specified, it will try to load the Broombridge data from the file
+        /// and will use the first ProblemDescription, otherwise, the problemDescription in the arguments is used.
+        /// </summary>
+        protected virtual CurrentVersion.ProblemDescription SelectProblemDescription(Arguments args)
```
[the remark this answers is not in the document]
anpaz-msft: I want to keep it protected virtual in case we want to create another magic that selects the problem description differently.

Chemistry/src/DataModel/Magic/FermionHamiltonianMagic.cs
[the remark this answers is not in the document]
anpaz-msft: We talked about it. For now, the Chemistry library is a C#-first and we will not attempt to make all its functionality available through Python.
There are a few methods that we will expose so we can use broombridge and Q# from Python, but we need to keep it minimal. For now these methods will require this round-trip. Eventually we might decide to implement them manually on Python.
Anything that requires building and manipulating Hamiltonians programmatically will have to be done in C#.

PR #67, opened 2019-04-11
Clean up json encoding for chemistry magic

anpaz-msft:


Chemistry/src/DataModel/LadderOperator/LadderOperator.cs
```diff
@@ -10,6 +11,7 @@ namespace Microsoft.Quantum.Chemistry.LadderOperators
     /// <summary>
     /// Data strcture for raising and lowering operators.
     /// </summary>
+    [JsonConverter(typeof(LadderOperator.JsonConverter))]
     public struct LadderOperator : IEquatable<LadderOperator>
```
guanghaolow: Note that LadderOperators are now generic in the update #66. However, only specializations LadderOperator<Int> (for current Hamiltonians) and LadderOperator<SpinOrbital> (for wavefunctions) are currently use.

Chemistry/src/DataModel/LadderOperator/LadderSequence.cs
```diff
@@ -4,13 +4,15 @@
...
 
 namespace Microsoft.Quantum.Chemistry.LadderOperators
 {
-    
+
     /// <summary>
     /// Class representing a sequence of raising and lowering operators.
     /// </summary>
+    [JsonConverter(typeof(LadderSequence.JsonConverter))]
     public class LadderSequence : IEquatable<LadderSequence>
```
guanghaolow: See [comment](https://github.com/Microsoft/QuantumLibraries/pull/67/files#r274252989)

Chemistry/src/DataModel/TermTypes.cs
```diff
@@ -88,6 +97,7 @@ public enum RaisingLowering
     /// Boxed version of double that implements the <see cref="ITermValue"></see> interface
     /// that requires all values to have a method to compute its norm.
     /// </summary>
```
guanghaolow: SpinOrbital.Indexconvention is moved to this file in the update

guanghaolow:
Looks good!

Chemistry/src/DataModel/LadderOperator/LadderOperator.cs
anpaz-msft: That might be problematic for this scenario, as I believe you can't specify a generic `JsonConverter<>`.
What we will need to do is to keep the `JsonConverter` generic but remove it from the base `LadderOperator` and then each concrete `LadderOperator` will have to specify its own `JsonConverter` attribute.

PR #74, opened 2019-04-12
Explicitly qualify all names in M.Q.Ext.Math.

cgranade:
This PR helps migrate to 0.6 namespacing by explicitly qualifying all names in the Microsoft.Quantum.Extensions.Math namespace. Once migration to 0.6 is complete, these qualifications can be subsquently removed.


cgranade:
Thanks for the review, @anpaz-msft!

PR #75, opened 2019-04-12
New array functions

cgranade:
This PR introduces two new array functions, `IndexRange` and `Enumerated`, for working with arrays and ranges in `for` loops.


anpaz-msft:
I wouldn't spend too much time figuring out why the linux build failed, seems like a transient linux-only problem

cgranade:
Thanks! I won't worry about the Linux build, then.

PR #88, opened 2019-04-22
Fix mutable array issue in chemistry data model.

cgranade:
This PR fixes the chemistry data model to not depend on mutability of Q# arrays at the C# level.


cgranade:
Thanks for the review, @guanghaolow!

PR #90, opened 2019-04-23
adding pytest runner files to make running tests easier

joesho112358:


cgranade:
While we appreciate the contribution, given how short the command to run `pytest` is, it would be more maintainable for us to modify the README to be more explicit about running tests than to add OS-dependent scripts. Given that, I'll go on and close this PR and make a new one to add that detail. Thank you for your suggestion as to how to improve the Quantum Development Kit!

PR #91, opened 2019-04-23
Continue migrating libraries to Q# 0.6

cgranade:


Chemistry/src/Runtime/JordanWigner/OptimizedBEOperator.qs
```diff
@@ -47,45 +47,44 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
...
+    function _OptimizedBEXY_ (targetIndex : Int) : ((Qubit, Qubit, Qubit[]) => Unit is Adj + Ctl) {
         //Message($"OptimizedBEXY {targetIndex}");
         return _OptimizedBEXY__(targetIndex, _, _, _);
     }
-    
-    
+
+
     // Subroutine of OptimizedBEXY.
     operation _OptimizedBEXY__ (targetIndex : Int, pauliBasis : Qubit, accumulator : Qubit, targetRegister : Qubit[]) : Unit {
```
bettinaheim: using a partial affiliation "is Adj" would be possible here

Chemistry/src/Runtime/JordanWigner/OptimizedBEOperator.qs
```diff
@@ -101,7 +100,7 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
     /// ## targetRegister
     /// Register of qubits on which the Pauli operators are applied.
     operation SelectZ (indexRegister : LittleEndian, targetRegister : Qubit[]) : Unit {
```
bettinaheim: is Adj + Ctl ?

Standard/src/Canon/Combinators/Deprecated.qs
```diff
@@ -2,44 +2,136 @@
...
-            Renamed("Microsoft.Quantum.Canon.WithCA", "Microsoft.Quantum.Canon.ApplyWithCA");
+            _Renamed("Microsoft.Quantum.Canon.WithCA", "Microsoft.Quantum.Canon.ApplyWithCA");
             ApplyWithCA(outerOperation, innerOperation, target);
         }
         adjoint auto;
         controlled auto;
         controlled adjoint auto;
     }
 
+    /// # Deprecated
```
bettinaheim: Do you want to move this into a separate file or something (like you did below)?

Chemistry/src/Runtime/JordanWigner/OptimizedBEOperator.qs
cgranade: Good catch, done.

Standard/src/Canon/Combinators/Deprecated.qs
cgranade: I think it should be OK here, since it's a deprecation stub for something originally in Combinators/*.qs. Ideally, the deprecation stub should go away shortly in any case.

PR #96, opened 2019-04-25
Begin adding more classical math functions for BigInt.

cgranade:
Still need to add unit tests.

### DIFF
--- a/Standard/src/Diagnostics/Facts.qs
+++ b/Standard/src/Diagnostics/Facts.qs
@@ -74,6 +74,25 @@ namespace Microsoft.Quantum.Diagnostics {
     }
 
     /// # Summary
+    /// Asserts that a classical BigInt variable has the expected value.
+    ///
+    /// # Input
+    /// ## actual
+    /// The number to be checked.
+    /// ## expected
+    /// The expected value.
+    ///
+    /// ## message
+    /// Failure message string to be used when the assertion is triggered.
+    function EqualityFactL(actual : BigInt, expected : BigInt, message : String) : Unit
+    {
+        if (actual != expected)
+        {
+            fail message;
+        }
+    }
+
+    /// # Summary
     /// Asserts that a classical Bool variable has the expected value.
     ///
     /// # Input
--- a/Standard/src/Math/Deprecated.qs
+++ b/Standard/src/Math/Deprecated.qs
@@ -8,14 +8,14 @@ namespace Microsoft.Quantum.Canon {
     /// # Deprecated
     /// This function has been removed.
     function IntAbs(input : Int) : Int {
-        // TODO: add call to _Removed.
+        _Removed("Microsoft.Quantum.Canon.IntAbs", "Microsoft.Quantum.Math.AbsI");
         return AbsI(input);
     }
 
     /// # Deprecated
     /// This function has been removed.
     function IntMax(a : Int, b : Int) : Int {
-        // TODO: add call to _Removed.
+        _Removed("Microsoft.Quantum.Canon.IntMax", "Microsoft.Quantum.Math.MaxI");
         return MaxI(a, b);
     }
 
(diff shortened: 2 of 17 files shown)


Standard/src/Diagnostics/Facts.qs
```diff
@@ -73,6 +73,25 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    ///
+    /// # Input
+    /// ## actual
+    /// The number to be checked.
+    /// ## expected
+    /// The expected value.
+    ///
+    /// ## message
+    /// Failure message string to be used when the assertion is triggered.
+    function EqualityFactL(actual : BigInt, expected : BigInt, message : String) : Unit
```
bettinaheim: My first thought was that this name is rather cryptic, then I remembered the style guide. 
Do we really want Asserts/Facts with such a narrow application? Why the need for a function (opposed to using e.g.  Fact(power >= 0, $"`power` must be non-negative"))?

Standard/src/Math/Deprecated.qs
```diff
@@ -25,4 +25,60 @@ namespace Microsoft.Quantum.Canon {
         _Renamed("Microsoft.Quantum.Canon.PNormalize", "Microsoft.Quantum.Math.PNormalized");
         return PNormalized(p, array);
     }
+
```
bettinaheim: Do you want to split out deprecations cleanly into separate files that can easily be removed when the time comes?

Standard/src/Math/Functions.qs
```diff
@@ -170,40 +164,49 @@ namespace Microsoft.Quantum.Math {
...
+        return (r < 0L) ? (r + modulus) | r;
     }
     
     
     /// # Summary
-    /// Let us denote expBase by x, power by p and modulus by N.
-    /// The function returns xᵖ mod N.
-    /// 
+    /// Let us denote expBase by $x$, power by $p$ and modulus by $N$.
+    /// The function returns $x^p \operatorname{mod} N$.
```
bettinaheim: Are we getting latex support for our docs? That would be super exciting!

Standard/src/Math/Functions.qs
```diff
@@ -250,23 +285,38 @@ namespace Microsoft.Quantum.Math {
...
+
+    /// # Summary
+    /// Internal recursive call to calculate the GCD.
+    function _ExtendedGreatestCommonDivisorL(signA : BigInt, signB : BigInt, r : (BigInt, BigInt), s : (BigInt, BigInt), t : (BigInt, BigInt)) : (BigInt, BigInt) {
+        if (Snd(r) == 0L) {
+            return (Fst(s) * signA, Fst(t) * signB);
+        }
+        
+        let quotient = Fst(r) / Snd(r);
+        let r_ = (Snd(r), Fst(r) - quotient * Snd(r));
```
bettinaheim: Lambda functions would make life much better here...

Standard/src/Math/Functions.qs
```diff
@@ -275,17 +325,56 @@ namespace Microsoft.Quantum.Math {
...
     /// # Output
-    /// Greatest common divisor of a and b
-    function GCD (a : Int, b : Int) : Int
-    {
-        let (u, v) = ExtendedGCD(a, b);
+    /// Tuple $(u,v)$ with the property $u \cdot a + v \cdot b = \operatorname{GCD}(a, b)$.
+    ///
+    /// # References
+    /// - This implementation is according to https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm
+    function ExtendedGreatestCommonDivisorL(a : BigInt, b : BigInt) : (BigInt, BigInt) {
```
bettinaheim: What is the reason for splitting this out into a separate function opposed to incorporating it into GreatestCommonDivisorL?

Standard/src/Math/Functions.qs
```diff
@@ -315,21 +404,64 @@ namespace Microsoft.Quantum.Math {
...
+        return _ContinuedFractionConvergentI(signA, signB, r, s, t, denominatorBound);
     }
     
+    /// # Summary
+    /// Internal recursive call to calculate the GCD with a bound
+    function _ContinuedFractionConvergentL(signA : BigInt, signB : BigInt, r : (BigInt, BigInt), s : (BigInt, BigInt), t : (BigInt, BigInt), denominatorBound : BigInt) : BigsFraction
+    {
+        if (Snd(r) == 0L or AbsL(Snd(s)) > denominatorBound)
+        {
+            if (Snd(r) == 0L and AbsL(Snd(s)) <= denominatorBound)
```
bettinaheim: the two returns could be unified into a single one using ?| (but that's a personal preference)

Standard/src/Math/Functions.qs
bettinaheim: Ah, I see it's used independently on GreatestCommonDivisorL. A don't know if there is a more descriptive name?

Standard/src/Math/Functions.qs
```diff
@@ -392,11 +558,42 @@ namespace Microsoft.Quantum.Math {
...
-        EqualityFactB(a >= 0, true, $"`a` must be non-negative");
+        Fact(a >= 0, $"`a` must be non-negative");
         return _bitsize(a, 0);
     }
+
+    /// # Summary
+    /// Helper function used to recursively calculate the bitsize of a value.
+    function _bitsize_l(val : BigInt, bitsize : Int) : Int
+    {
+        if (val == 0L)
```
bettinaheim: this now definitively feels to me like it should be unified into a single return statement.

cgranade: Good point, yeah.

Standard/src/Math/Functions.qs
cgranade: "Extended GCD" is a fairly common name, but I am definitely open to suggestions.

Standard/src/Math/Functions.qs
cgranade: Absolutely.

Standard/src/Math/Deprecated.qs
cgranade: Definitely, my strategy so far has been to make sure all deprecation stubs live in files named `Deprecated.qs`.

Standard/src/Diagnostics/Facts.qs
cgranade: It seems to be a fairly common unit testing package idiom to offer equality checks and to typeset what the expected and actual value automatically; we'd already provided that idiom through `AssertIntEqual` (now `EqualityFactI`), so I was following that pattern here.

Standard/src/Math/Functions.qs
cgranade: My understanding is that the docs at docs.microsoft.com/quantum and docs.microsoft.com/qsharp support LaTeX, but that best practice is to avoid math at all in summaries (saving it for the description section). I'll rework accordingly.

Standard/src/Math/Functions.qs
cgranade: My preference as well; I only did that here in keeping with old (pre `?|`) code, but I've updated now. Thanks for catching that!

Standard/src/Diagnostics/Facts.qs
bettinaheim: Yes, I know. I never quite understood the need for that - just a thought though.

Standard/src/Math/Functions.qs
bettinaheim: I can't come up with anything better either.

Standard/src/Diagnostics/Facts.qs
cgranade: Fair enough! `AssertIntEqual` was widely enough used that I didn't feel like it was great to strip it out yet, but I do think that at some point we need to take a coherent look at unit testing in Q# as a whole. Deferring for now, though, if that's OK.

Standard/src/Math/Functions.qs
cgranade: I'll leave it for now, then, if that's OK? It's already a more descriptive name than before, at least...

bettinaheim: Sure.

PR #100, opened 2019-05-08
Update release/v0.7 to v0.6 syntax

guanghaolow:


Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -60,7 +62,7 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
         for(idx in 0..Length(idxFermions)-1){
```
cgranade: This can now be replaced with `idx in IndexRange(idxFermions)`.

anpaz-msft: done.

anpaz-msft:
The [WARNING] are all existing from 0.6, so I'll send a different PR to fix them. We'll fix the `SystemTests` later.

guanghaolow:
@anpaz-msft Some of the `SystemTests` use phase estimation and are probabilistic, which explains the random failures. How does one go about fixing a seed for each test?

anpaz-msft:
@guanghaolow, we run the tests using a seed for the rng, so that way the test results become deterministic. Because the tests kept failing, I've added the seed to the LiH tests: see [commit f9cc661](https://github.com/microsoft/QuantumLibraries/pull/100/commits/f9cc661fb833fcc9cd6998c83c14ee5fdd03b6bd)

Issue #34, opened 2019-02-25
Higher-level numerics functionality


cgranade:
Many quantum algorithms like HHL rely on having higher level numeric operations available (e.g.: 1 / x, arcsin, etc.). It would be great to add operations supporting these applications to the Quantum Development Kit's standard libraries, building on existing arithmetic work and the enhancements under #33.

cgranade:
Forgot to close this one when #102 landed. Closing now.

PR #102, opened 2019-05-14
Add more numerics functionality

thomashaener:
Addressing #34 by adding more numerics functionality (integer and fixed-point arithmetic).


Numerics/tests/FixedPointTests.qs
```diff
@@ -0,0 +1,396 @@
...
+
+    operation InitFxPTest() : Unit {
+        for (a in [1.2, 3.9, 3.14159, -0.6, -4.5, -3.1931, 0.0]){
+            using (xs = Qubit[10]) {
+                let fp = FixedPoint(4, xs);
+                InitFxP(a, fp);
+                let measured = MeasureFxP(fp);
+                EqualityFactB(AbsD(measured - a) <= 1./IntAsDouble(2^7), true,
+                    $"FixedPoint initialized to {a} but measured {measured}.");
+                ResetAll(xs);
```
cgranade: Since `MeasureFxP` also resets, it might be good to remove this call so that this unit test also checks that `MeasureFxP` does in fact reset its qubits. If you do so, I'd recommend a comment to that effect so that if the test fails, it's not so mysterious.

Numerics/tests/FixedPointTests.qs
```diff
@@ -0,0 +1,396 @@
...
+
+    operation AddConstantFxPTest() : Unit {
+        for (a in [1.2, 3.9, 3.14159, -0.6, -4.5, -3.1931, 0.0]){
+            for (b in [1.2, 3.9, 3.14159, -0.6, -4.5, -3.1931, 0.0]){
+                using (xs = Qubit[11]) {
+                    let fp = FixedPoint(5, xs);
+                    InitFxP(a, fp);
+                    AddConstantFxP(b, fp);
+                    let measured = MeasureFxP(fp);
+                    EqualityFactB(AbsD(measured - (a+b)) <= 1./IntAsDouble(2^6),
```
cgranade: This looks like another good usecase for EqualityWithinToleranceFact.

Numerics/src/FixedPoint/Addition.qs
```diff
@@ -0,0 +1,57 @@
...
+    /// # Summary
+    /// Addition of a classical constant to a quantum fixed-point number
+    ///
+    /// # Input
+    /// ## constant
+    /// Constant to add to the quantum fixed-point number.
+    /// ## fp
+    /// Fixed-point number (of type FixedPoint), to which the constant will
+    /// be added.
+    operation AddConstantFxP(constant : Double, fp : FixedPoint) : Unit {
```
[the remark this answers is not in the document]
cgranade: After discussion, let's keep `AddConstant` for now and perhaps change `IncrementBy` later to match.

PR #108, opened 2019-05-21
Add user agent and mock client to Python.

cgranade:


Interoperability/python/src/qsharp/clients/iqsharp.py
```diff
@@ -91,7 +91,7 @@ def busy(self) -> bool:
         return self._busy
 
     def compile(self, body):
-        return self._execute(body)
+        return "E.F.G"
```
anpaz-msft: I think this slipped away.

Interoperability/python/src/qsharp/clients/mock.py
```diff
@@ -0,0 +1,99 @@
...
+from qsharp.serialization import map_tuples, unmap_tuples
+
+## LOGGING ##
+
+import logging
+logger = logging.getLogger(__name__)
+
+## CLASSES ##
+
+class MockClient(object):
```
anpaz-msft: It is probably useful to add a `logger.info` or `logger.debug` for every command.

Interoperability/python/src/qsharp/clients/mock.py
```diff
@@ -0,0 +1,99 @@
...
+# -*- coding: utf-8 -*-
+##
+# mock.py: Placeholder client useful in running unit tests and in hosted CI.
+##
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+##
+
+
+import subprocess
```
anpaz-msft: most if not all of these imports are not needed.

Interoperability/python/src/qsharp/clients/iqsharp.py
cgranade: Yes, good catch! I'll fix it, then.

Interoperability/python/src/qsharp/clients/mock.py
cgranade: I like that idea, yeah. I'll go on and implement it then.

Interoperability/python/src/qsharp/clients/mock.py
cgranade: Right, I forgot to clean that up, sorry.

PR #110, opened 2019-05-24
Add user agent and mock client to Python. (#108)

anpaz-msft:
* Add user agent and mock client to Python.


anpaz-msft:
Cherry picking python's user-agent change into master.

PR #112, opened 2019-05-24
A couple minor fixes to Facts api docs.

cgranade:


tcNickolas:
Hmm, do we have a flaky test in the libraries? This change doesn't look like it should break anything...

cgranade:
That's odd, the tests didn't even fail, but the build itself was aborted on macOS?

cgranade:
I triggered a new build in case the macOS agent itself is being flaky.

PR #114, opened 2019-05-25
Clean up chemistry.

anpaz-msft:
Fixing a bug that was ignoring the `IndexConvention` argument in the `FermionHamiltonian` Magic.
Removed chemistry samples files are they are now in the `/Quantum` repository.


Chemistry/src/Jupyter/FermionHamiltonianMagic.cs
```diff
@@ -78,7 +78,7 @@ public ExecutionResult Run(string input, IChannel channel)
             // We can obtain the full fermion Hamiltonian from the more compact orbital integral representation.
             // This transformation requires us to pick a convention for converting a spin-orbital index to a single integer.
             // Let us pick one according to the formula `integer = 2 * orbitalIndex + spinIndex`.
```
cgranade: Is this comment still accurate?

PR #118, opened 2019-05-29
Added new Apply{,A,C,CA} operation.

cgranade:


cgranade:
Thanks for the review! This was added to address a usecase within the
katas, where there's a loop that passes each of an array of qubits to each
of an array of operations (as in ops[0](qs[0]), ops[1](qs[1]), and so
forth).

PR #128, opened 2019-06-20
Minor improvements to Python docstrings.

cgranade:


Interoperability/python/src/qsharp/types.py
```diff
@@ -18,25 +18,43 @@
 ## ENUMS ######################################################################
 
 class Result(IntEnum):
+    """
+    Represents the `Result Q# type <https://docs.microsoft.com/quantum/language/type-model#primitive-types>`_.
```
tcNickolas: I'm not familiar enough with Python docstrings to figure out what the underscore in the end is supposed to do, and a quick search through documentation didn't help.

cgranade: Python docstrings are generally written in the [reStructuredText language](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) (it predates Markdown, and there's never been enough impetuous to switch over). In RST, the notation `` `title <url>`_ `` is used to indicate an inline link.

cgranade:
Thanks for the review, @tcNickolas! I'll go on and merge this in once CI passes with updates from master.

PR #130, opened 2019-06-21
Added metadata to libraries as well.

cgranade:
This PR parallels microsoft/Quantum#201 by adding a contribution guide and issue templates to the Quantum Development Kit libraries repo as well.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is, and which sample or samples you encountered the bug in.
```
tcNickolas: This phrasing is a bit confusing in the context of libraries - the issues with library code not necessarily manifest in samples, they might manifest in user code, in katas etc. I'd suggest something like "the code that can be used to reproduce the bug (one of the Quantum Development Kit samples, Quantum Katas or your code)", possibly with links to samples and katas.

The same concern applies to line 35 of `CONTRIBUTING.md`

cgranade: Fixed, thanks for the suggestion.

Issue #104, opened 2019-05-18
Provide more array initialization functions, range conversions


cgranade:
As raised by @bettinaheim on Microsoft/Quantum#182, we should build on the enhancements in #39 by providing more functions to prepare arrays.
For instance, the ReversibleLogicSynthesis sample contributed by @msoeken implements two functions that are good candidates for Microsoft.Quantum.Arrays:

- `Sequence : (Int, Int) -> Int[]`: returns an array of integers in the given (inclusive) interval.
- `Numbers : Int -> Int[]`: returns an array of integers in the given half-open interval starting at 0.

Implementing these and other similar array initialization functions would allow us to better isolate mutability and provide more reusable conveniences to users.

PR #140, opened 2019-07-19
Array initialization functions and range conversions

msoeken:
This PR implements, as described in #104, several array initialization and range conversion functions:

* `SequenceI (from : Int, to : Int) : Int[]` Array of integers `from` to `to`
* `SequenceL (from : BigInt, length : Int) : BigInt[]` Array of integers from `from` of length `length`
* `Numbers (count : Int) : Int[]` Array of integers from `0` to `count - 1`
* `ArrayFromRange (range: Range) : Int[]` Array with elements visited in `range`

**Remarks:**

* There is an inconsistency in the parameters in `SequenceI` and `SequenceL`, since the size of an array is initialized using an `Int`.
* The function name `Numbers` may be too general.
* The function `ArrayFromRange` iterates the range twice. For simple ranges with increment `1`, SequenceI is faster. This is noted in the doc string.


cgranade:
Thanks for the PR, @msoeken! I'll review momentarily, then. In the meantime, with respect to your remarks, had just a few thoughts:

> **Remarks:**
> 
> * There is an inconsistency in the parameters in `SequenceI` and `SequenceL`, since the size of an array is initialized using an `Int`.

I think it could be more natural here to keep the inputs the same (other than type), and to fail if `from - to` is too large to store in an `Int`.

> * The function name `Numbers` may be too general.

Agreed, perhaps it may be better to use `SequenceI(0, _)` in preference to `Numbers`?

> * The function `ArrayFromRange` iterates the range twice. For simple ranges with increment `1`, SequenceI is faster. This is noted in the doc string.

We currently have [`RangeAsIntArray](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.convert.rangeasintarray), which uses the [`RangeStart`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.core.rangestart), [`RangeEnd`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.core.rangeend, and [`RangeStep`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.core.rangestep) functions from Microsoft.Quantum.Core to avoid looping through twice, perhaps it would be good to use that approach here as well?

Thanks again for the PR, really appreciate it!

msoeken:
Thanks for your comments, @cgranade.

1. I have unified `SequenceL` with `SequenceI` by providing to both `from` and `to` parameters. Right now there is no `BigIntAsInt` function in Q#, which probably needs to be implemented in *qsharp-runtime*. I have a work-around through an `Bool[]`, but this seems to be an overkill. Should I make a issue or PR to integrate `BigIntAsInt` into *qsharp-runtime* first before proceedings. Or is there another way that I am overlooking?
2. I removed `Numbers` and mention in the docs to `SequenceI` that one can use it to generate a numbers function in the way you suggested it. However, the semantics is a little different. For example, before `Numbers(5)` did not include `5` but now it does.
3. I used the Range functions to implement my private `_RangeLength` function. This may be a good candidate for `Microsoft.Quantum.Core` as well.

Another question came to mind though: Would it make sense to have `ArrayFromRange` in the `Microsoft.Quantum.Convert` namespace instead of `Microsoft.Quantum.Arrays`?

cgranade:
> 1. I have unified `SequenceL` with `SequenceI` by providing to both `from` and `to` parameters. Right now there is no `BigIntAsInt` function in Q#, which probably needs to be implemented in _qsharp-runtime_. I have a work-around through an `Bool[]`, but this seems to be an overkill. Should I make a issue or PR to integrate `BigIntAsInt` into _qsharp-runtime_ first before proceedings. Or is there another way that I am overlooking?

Oh, goodness, that is a good point. We've generally only used `XAsY` functions for type conversions that always succeed, but that presents a bit of a problem for `BigIntAsInt`. I'm not sure what the best name would be, but it definitely seems like we will want something to drop back down to ints from big ints in qsharp-runtime.

> 2. I removed `Numbers` and mention in the docs to `SequenceI` that one can use it to generate a numbers function in the way you suggested it. However, the semantics is a little different. For example, before `Numbers(5)` did not include `5` but now it does.

That's a good point, but I think it still is worth it to reduce duplication in the API offered by the standard libraries.

> 3. I used the Range functions to implement my private `_RangeLength` function. This may be a good candidate for `Microsoft.Quantum.Core` as well.

`RangeLength` might be a good candidate for inclusion in Microsoft.Quantum.Core, though we may also want to take the whole `Range` type in another direction as well. At one point, @bettinaheim and I had floated the idea of making `Range` a UDT so that we don't need the `Start`, `Step`, and `End` functions. Doing that, `Length` could make sense as a specialization `Length<Range>`, depending on how that language feature comes together.

> Another question came to mind though: Would it make sense to have `ArrayFromRange` in the `Microsoft.Quantum.Convert` namespace instead of `Microsoft.Quantum.Arrays`?

It might make sense, yeah. That was the thinking behind [`RangeAsIntArray`](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.convert.rangeasintarray), for instance.  Maybe it would it make sense to consolidate `ArrayFromRange` with `RangeAsIntArray`, improving the implementation and documentation of the latter?

msoeken:
Hi @cgranade, thanks for your comments and answers.

> > 1. I have unified `SequenceL` with `SequenceI` by providing to both `from` and `to` parameters. Right now there is no `BigIntAsInt` function in Q#, which probably needs to be implemented in _qsharp-runtime_. I have a work-around through an `Bool[]`, but this seems to be an overkill. Should I make a issue or PR to integrate `BigIntAsInt` into _qsharp-runtime_ first before proceedings. Or is there another way that I am overlooking?
> 
> Oh, goodness, that is a good point. We've generally only used `XAsY` functions for type conversions that always succeed, but that presents a bit of a problem for `BigIntAsInt`. I'm not sure what the best name would be, but it definitely seems like we will want something to drop back down to ints from big ints in qsharp-runtime.
> 

Is it worth to move this discussion to *qsharp-runtime*, maybe with a PR? It's little work.

> > 1. I removed `Numbers` and mention in the docs to `SequenceI` that one can use it to generate a numbers function in the way you suggested it. However, the semantics is a little different. For example, before `Numbers(5)` did not include `5` but now it does.
> 
> That's a good point, but I think it still is worth it to reduce duplication in the API offered by the standard libraries.
> 

Yes, I agree.

> > 1. I used the Range functions to implement my private `_RangeLength` function. This may be a good candidate for `Microsoft.Quantum.Core` as well.
> 
> `RangeLength` might be a good candidate for inclusion in Microsoft.Quantum.Core, though we may also want to take the whole `Range` type in another direction as well. At one point, @bettinaheim and I had floated the idea of making `Range` a UDT so that we don't need the `Start`, `Step`, and `End` functions. Doing that, `Length` could make sense as a specialization `Length<Range>`, depending on how that language feature comes together.
> 
> > Another question came to mind though: Would it make sense to have `ArrayFromRange` in the `Microsoft.Quantum.Convert` namespace instead of `Microsoft.Quantum.Arrays`?
> 
> It might make sense, yeah. That was the thinking behind [`RangeAsIntArray`](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.convert.rangeasintarray), for instance. Maybe it would it make sense to consolidate `ArrayFromRange` with `RangeAsIntArray`, improving the implementation and documentation of the latter?

I'll implement it as `RangeAsIntArray` in `Microsoft.Quantum.Convert` and leave `_RangeLength` a private function for now.

cgranade:
> Hi @cgranade, thanks for your comments and answers.

Thanks for the PR, and for the discussion!
 
> Is it worth to move this discussion to _qsharp-runtime_, maybe with a PR? It's little work.

Probably, given that any BigInt → Int function would have to be intrinsic anyway.

> > 
> > It might make sense, yeah. That was the thinking behind [`RangeAsIntArray`](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.convert.rangeasintarray), for instance. Maybe it would it make sense to consolidate `ArrayFromRange` with `RangeAsIntArray`, improving the implementation and documentation of the latter?
> 
> I'll implement it as `RangeAsIntArray` in `Microsoft.Quantum.Convert` and leave `_RangeLength` a private function for now.

Sounds good!

msoeken:
I noticed that `RangeAsIntArray` already exactly does the exact same thing as `ArrayFromRange`, so I removed it from the PR. I have also opened https://github.com/microsoft/qsharp-runtime/pull/33 to discuss about the conversion from `BigInteger` to `Int`.

Issue #142, opened 2019-07-31
Variants of ApplyIf that use Result type instead of Bool


vadym-kl:
**Is your feature request related to a problem? Please describe.**
Resource estimation of code involving measurements is a challenge for `ResourceEstimator` Q# simulator. Currently the only way to help `ResourceEstimator` to deal with measurement is to use `AssertProb` to provide measurement outcome probability. There are resource estimation scenarios where this is not sufficient.

**Describe the solution you'd like**
Adding combinators `ApplyIfOne`, `ApplyIfZero`, `ApplyIfElse` to `Microsoft.Quantum.Canon` will help to express patterns of code that `ResourceEstimator` has potential to handle. 
Suggested signatures are the following ( +usual `A, CA, C` versions) : 

```qsharp
operation ApplyIfOne<'T>( 
    measurementResult : Result, 
    (onResultOneOp : ('T => Unit), oneArg : 'T) 
) : Unit

operation ApplyIfZero<'T>( 
    measurementResult : Result,
    (onResultZeroOp : ('T => Unit), zeroArg : 'T) 
) : Unit

operation ApplyIfElse<'T,'U>(
    measurementResult : Result,
    (onResultZeroOp : ('T => Unit), zeroArg : 'T) ,
    (onResultOneOp : ('U => Unit), oneArg : 'U) 
) : Unit

// Example of calls: 
ApplyIfOne(res, (CX,(q1, q2)) ); 
// note how expression (CX, (q1, q2)) is reminiscent to usual call CX(q1,q2)
ApplyIfElse(res, (CX,(q1, q2)), (X,q3) );
```

**Describe alternatives you've considered**
One alternative is to replace type of `onResultZeroOp` and others by `Unit => Unit` and use 
```
operation HoldOperation<'T> ( op : ('T => Unit), arg : 'T, dummy : Unit ) : Unit {
  op(arg);
}
```
to reduce general case to `Unit => Unit`, but it seems to produce cumbersome code and has potential run-time overhead. 

**Additional context**
This is a prerequisite for potential improvements to `QCTraceSimulator` and `ResourceEstimator`. 
Feedback from Q# community and from @cgranade in particular about the choice of names and signatures is very welcome.

vadym-kl:
Related PR #143

cgranade:
Resolved with #143.

PR #143, opened 2019-07-31
Adding new variants of ApplyIf combinator into Microsoft.Quantum.Canon 

vadym-kl:
Work related to #142


Standard/src/Canon/Combinators/CControlled.qs
```diff
@@ -132,7 +132,364 @@ namespace Microsoft.Quantum.Canon
         controlled distribute;
         controlled adjoint distribute;
     }
+
+    /// # Summary
+    /// Given an operation `op` and a Result value `result`, applies `op` to the `target`
```
cgranade: Similarly as with #145, I would recommend minimal to no formatting in `# Summary` sections, due to limitations with how language server clients (e.g.: VS and VS Code) handle formatting in hover information.

Standard/src/Canon/Combinators/CControlled.qs
```diff
@@ -132,7 +132,364 @@ namespace Microsoft.Quantum.Canon
...
+    ///
+    /// # Type Parameters
+    /// ## 'T
+    /// The input type of the operation to be conditionally applied.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplyIfOneC
+    /// - Microsoft.Quantum.Canon.ApplyIfOneA
+    /// - Microsoft.Quantum.Canon.ApplyIfOneCA
+    operation ApplyIfOne<'T> (result : Result, (op : ('T => Unit), target : 'T)) : Unit
```
cgranade: Very minor, but Q# style has generally been to include braces in the same line (that is, TypeScript-style rather than C#-style).

Standard/src/Canon/Combinators/CControlled.qs
```diff
@@ -132,7 +132,364 @@ namespace Microsoft.Quantum.Canon
...
+
+    /// # Summary
+    /// Given an operation `op` and a Result value `result`, applies `op` to the `target`
+    /// if `result` is One. If result is Zero, nothing happens to the `target`.
+    ///
+    /// # Input
+    /// ## op
+    /// An operation to be conditionally applied.
+    /// ## result
+    /// a measurement result that controls whether op is applied or not.
```
cgranade:
```suggestion
    /// A measurement result that controls whether op is applied or not.
```

Standard/src/Canon/Combinators/CControlled.qs
```diff
@@ -132,7 +132,364 @@ namespace Microsoft.Quantum.Canon
...
+    /// ## 'T
+    /// The input type of the operation `opZero` to be conditionally applied.
+    /// ## 'U
+    /// The input type of the operation `opOne` to be conditionally applied.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplyIfElseC
+    /// - Microsoft.Quantum.Canon.ApplyIfElseA
+    /// - Microsoft.Quantum.Canon.ApplyIfElseCA
+    operation ApplyIfElse<'T,'U> (result : Result, (opZero : ('T => Unit), targetZero : 'T), (opOne : ('U => Unit), targetOne : 'U)) : Unit
```
cgranade: I think the name here could be a bit confusing, as `ApplyIf` takes a `Bool` as input, while `ApplyIfElse` takes its input as a `Result`. That might could be resolved by adding `B` and `R` suffixes, but then that raises why `ApplyIfElse` needs a suffix while `ApplyIf` doesn't. We may want to discuss this one a bit more; in particular, I'd like to understand a bit about how this will likely be used in contrast to calling `ApplyIfZero(result, opZero, targetZero)` and then `ApplyIfOne(result, opOne, targetOne)`, so that I can help suggest a way to resolve the contradiction with `ApplyIf`.

vadym-kl: I think `ApplyIfElse ` is just more convenient alternative to 
```
ApplyIfZero(result, opZero, targetZero);
ApplyIfOne(result, opOne, targetOne);
```
I am OK with adding `R` suffix and also introducing `ApplyIfElseB` for Boolean type for completeness.

@cgranade , does this sound reasonable? 

Thanks for feedback.

Standard/src/Canon/Combinators/CControlled.qs
vadym-kl: Removed formatting from summary

Standard/src/Canon/Combinators/CControlled.qs
vadym-kl: Fixed the braces

Standard/src/Canon/Combinators/CControlled.qs
vadym-kl: Added `ApplyIfElseB` and `ApplyIfElseR` with variants.

PR #153, opened 2019-08-28
Add repeated operation

christopherkang:
Draft! Adding the following ops:

- ApplySeriesOfOps (and C/A/CA functor versions) - apply a list of ops and their targets
- ApplyOpRepeatedlyOver (and C/A/CA functor versions) - apply a single op multiple times with different targets
- PermuteQubits - uses ApplyOpRepeatedlyOver to SWAP over qubit array and permute their order.

Would like some input on a few things:

- Do the operation signatures make sense? Are there operations that could be instead written as functions?
- Is the Q# file in the appropriate place? 
- Anything I'm missing? :)

To-do:

- [x] Add unit tests
- [x] Finish docs
- [x] Finish C/A/CA functor versions of ApplySeriesOfOps


christopherkang:
@msoeken is this achievable without mutating an array? It would be preferable to have this op as Adj+Ctl.

cgranade:
@christopherkang: One trick is to isolate mutability within a function, then call to that function from within an operation with `is Adj + Ctl`. There's a number of existing functions in Microsoft.Quantum.Arrays for just this reason, for instance.

christopherkang:
@cgranade I used that trick to implement the function. @msoeken I'll start working on the unit tests - do you have any suggestions on particular scenarios to use to test the functions? I'd appreciate input from you both on the additions made.

msoeken:
Hi @christopherkang, thanks for the additions. I have worked on a code for permutation synthesis some weeks ago, but it hasn't made it yet to the libraries. There is no need to have two functions that do the same thing, so I'd like to share my code with you as it may be helpful for your implementation.

```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

namespace Microsoft.Quantum.Canon {
    open Microsoft.Quantum.Arrays;
    open Microsoft.Quantum.Convert;

    function _Equals(a : Int, b : Int) : Bool {
        return a == b;
    }

    function _NotEquals(a : Int, b : Int) : Bool {
        return a != b;
    }

    function _IsPermutationPred(permutation : Int[], value : Int) : Bool {
        let index = IndexOf(_Equals(value, _), permutation);
        return index != -1;
    }

    function _IsPermutation(permutation : Int[]) : Bool {
        return All(_IsPermutationPred(permutation, _), RangeAsIntArray(IndexRange(permutation)));
    }

    function _PermutationToTransposition(permutation : Int[]) : (Int, Int)[] {
        Fact(_IsPermutation(permutation), $"`permutation` is not a permutation");

        mutable preTrans = new (Int, Int)[Length(permutation) - 1];
        mutable p = permutation;

        for (i in Length(p) - 1..-1..1) {
            let j = IndexOf(_Equals(i, _), p);
            if (i != j) {
                set p w/= j <- p[i];
                set p w/= i <- i;
                set preTrans w/= Length(preTrans) - i <- (i, j);
            } else {
                set preTrans w/= Length(preTrans) - i <- (i, i);
            }
        }

        return Filtered(_NotEquals, preTrans);
    }

    /// # Summary
    /// Explicitly permutes qubits using SWAP gates subject to given
    /// permutation.
    ///
    /// # Input
    /// ## qs
    /// Array of qubits to permute.
    ///
    /// ## permutation
    /// Permutation of array `[0, 1, ..., n - 1]`, where `n - 1` is the length
    /// of array `qs`.
    ///
    /// # Remarks
    /// The lengths of `qs` and `permutation` must be equal.  Also `permutation`
    /// must be a correct permutation, i.e., it contains all elements from `0`
    /// to `n - 1` exactly once.
    operation PermuteQubits(qs : Qubit[], permutation : Int[]) : Unit is Adj + Ctl {
        // check that sizes match
        EqualityFactI(Length(qs), Length(permutation), $"Lengths of `qs` and `permutation` must be equal");

        for ((i, j) in _PermutationToTransposition(permutation)) {
            SWAP(qs[i], qs[j]);
        }
    }
}
```

msoeken:
Also note that functions such as `ClaimEqualInt` or `_Equals` and `_NotEquals` in your code are currently addressed in https://github.com/microsoft/QuantumLibraries/pull/152, so that PR can make the code a bit simpler, once it has been merged.

christopherkang:
@msoeken I updated my code with your suggestions. I've kept claims in for now - after Quantum.Logical gets pushed, I'll update the code with the new methods

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+        return All(_IsPermutationPred(permuation, _), RangeAsIntArray(IndexRange(permuation)));
+    }
+
+    /// # Summary
+    /// Returns the order elements in an array need to be swapped to produce an ordered array. 
+    /// Assumes swaps occur in place.
+    ///
+    /// # Input
+    /// ## newOrder
+    /// Array with the permutation of the indices of the new array. There should be n elements,
```
cgranade: I'd suggest either using TeX notation for `n` (`` $n$ ``), or typesetting as a Q# variable (`` `n` ``).

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+    ///     {
+    ///         Switch newOrder[index] with newOrder[newOrder[index]]
+    ///     }
+    /// }
+    function SwapOrderToPermuteArray(newOrder : Int[]) : (Int, Int)[] {
+        // Check to verify the new ordering actually is a permutation of the indices
+        Fact(_IsPermutation(newOrder), $"The new ordering is not a permutation");
+
+        // The maximum number of swaps is n - 1
+        mutable swaps = new (Int, Int)[Length(newOrder) - 1];
```
cgranade: I'd suggest making this a length-0 array, then appending to it from the loop below. If needed, you can add a fact to check that the length is less than or equal to `n - 1`. That avoids needing to track `swapIndex`, and to filter at the end.

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// Index of the second element to be swapped.
+    ///
+    /// ## arr
+    /// Array with elements to be swapped.
+    ///
+    /// # Output
+    /// The array with the in place swapp applied.
+    function Swapped<'T>(firstIndex: Int, secondIndex: Int, arr: 'T[]) : 'T[] {
+        mutable newArray = arr;
+        let firstIndexVal = arr[firstIndex];
```
cgranade: You can actually implement this without mutability at all by chaining the `w/` operator twice.
```suggestion
        return arr
            w/ firstIndex <- arr[secondIndex]
            w/ secondIndex <- arr[firstIndex];
```

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,236 @@
...
+    /// ## listOfOps
+    /// List of ops, each taking a qubit array, to be applied. They are applied sequentially, lowest index first.
+    /// ## targets
+    /// Nested arrays describing the targets of the op. Each array should contain a list of ints describing 
+    /// the qubits to be used.
+    /// ## register
+    /// Qubit register to be acted upon.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplyOpRepeatedlyOver
```
cgranade: An example here would be great, I think.

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,236 @@
...
+    operation ApplySeriesOfOps(listOfOps : (Qubit[] => Unit)[], targets : Int[][], register : Qubit[]) : Unit {
```
cgranade: This readily generalizes to arbitrary types.
```suggestion
    operation ApplySeriesOfOps<'T>(listOfOps : ('T[] => Unit)[], targets : Int[][], register : 'T[]) : Unit {
```

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,236 @@
...
+        if (Length(listOfOps) != Length(targets)) {
+            fail "The number of ops and number of targets do not match!";
+        }
+        for (index in 0..Length(listOfOps) - 1) {
```
cgranade: Since you never directly use the value of `index`, I'd suggest looping over the elements of interest instead of their indices.
```suggestion
        for ((op, targetIndices) in Zip(listOfOps, targets)) {
```

cgranade: More radically, the entire loop could be collapsed into a single call:
```Q#
ApplyToEach(
     Delay(_, _, ()),
     Zip(
          listOfOps,
          Mapped(
              Subarray(_, register),
              targets
          )
     )
);
```
There are definitely valid reasons to not do so, but I wanted to at least highlight the options available here.

Standard/src/Canon/CommonGates.qs
```diff
@@ -272,4 +274,20 @@ namespace Microsoft.Quantum.Canon {
         CZ(qubit1, qubit2);
     }
 
+    /// # Summary
+    /// Permutes qubits by using the SWAP operator.
```
cgranade:
```suggestion
    /// Permutes qubits by using the SWAP operation.
```

Standard/src/Canon/CommonGates.qs
```diff
@@ -272,4 +274,20 @@ namespace Microsoft.Quantum.Canon {
...
+    ///
+    /// # Input
+    /// ## ordering
+    /// Describes the new ordering of the qubits, where the qubit at index i will now be at ordering[i].
+    /// ## register
+    /// Qubit register to be acted upon.
```
cgranade: An example giving two equivalent Q# snippets would be really helpful here.

Standard/src/Standard.csproj
```diff
@@ -7,7 +7,7 @@
 
   <PropertyGroup>
     <NoWarn>0162</NoWarn>
-    <RunQDocGen>True</RunQDocGen>
+    <RunQDocGen>False</RunQDocGen>
```
cgranade: This is fine for local testing, but should be reverted as a part of the PR.

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,236 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Canon {
+    open Microsoft.Quantum.Diagnostics;
+    open Microsoft.Quantum.Arrays;
+    open Microsoft.Quantum.Intrinsic;
+    
+    ///////////////////////////////////////////////////////////////////////////////////////////////
+    // Helpers to repeatedly apply functions over qubit arrays
```
cgranade:
```suggestion
    // Helpers to repeatedly apply operations over qubit arrays
```

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+                set swapIndex = swapIndex + 1;
+            }
+        }
+
+        // Remove (0, 0) swaps at the end
+        return Filtered(ClaimDifferentInt, swaps);
+    }
+
+    /// # Summary
+    /// Applies an in-place swap of two elements in an array.
```
cgranade: It's not actually in-place, since arrays are passed by value in Q#, not by reference.

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+    ///
+    /// ## secondIndex
+    /// Index of the second element to be swapped.
+    ///
+    /// ## arr
+    /// Array with elements to be swapped.
+    ///
+    /// # Output
+    /// The array with the in place swapp applied.
+    function Swapped<'T>(firstIndex: Int, secondIndex: Int, arr: 'T[]) : 'T[] {
```
msoeken: It would be great if you can add a small example on how to use the function, if it ends up in the API.

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// # Summary
+    /// Turns a list of 2-tuples into a nested array.
+    ///
+    /// # Input
+    /// ## tupleList
+    /// List of 2-tuples to be turned into a nested array.
+    ///
+    /// # Output
+    /// A nested array with length matching the tupleList.
+    function TupleListToNestedArray<'T>(tupleList : ('T, 'T)[]) : 'T[][] {
```
msoeken: Here, an example would be very helpful. The description is quite abstract.

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,236 @@
...
+            if (Length(targets[index]) > Length(register)) 
+            {
+                fail "Too many targets!";
+            }
+            let opTargets = Subarray(targets[index], register);
+            op(opTargets);
+        }
+    }
+
+}
```
msoeken: Please add a new line to the end of the file.

Standard/tests/ApplyRepeatedOverTests.qs
```diff
@@ -0,0 +1,67 @@
...
+    operation _SamplePermuteQubits(register : Qubit[]) : Unit is Adj + Ctl {
+        // assumes the order to be swapped is [(0, 5),(0, 4),(0, 1),(0, 3)]
+        // (Order is [5, 3, 2, 0, 1, 4])
+        SWAP(register[0], register[5]);
+        SWAP(register[0], register[4]);
+        SWAP(register[0], register[1]);
+        SWAP(register[0], register[3]);
+    }
+    
+}
```
msoeken: Please add a new line to the end of the file

Standard/src/Arrays/Arrays.qs
christopherkang: What's the syntax for appending to an array?

cgranade: The `+` operator concatenates arrays, and can even be used with `set ... += ...` statements. Thus, you can do something like `set swaps += [(index, order[index])];` inside your loop.

Standard/src/Standard.csproj
christopherkang: Is there a way to build locally with this set to True? My build / tests fail otherwise

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +282,105 @@ namespace Microsoft.Quantum.Arrays {
...
+    ///
+    /// ## Psuedocode
+    /// for (index in 0..Length(newOrder) - 1) 
+    /// {
+    ///     while newOrder[index] != index 
+    ///     {
+    ///         Switch newOrder[index] with newOrder[newOrder[index]]
+    ///     }
+    /// }
+    function SwapOrderToPermuteArray(newOrder : Int[]) : (Int, Int)[] {
```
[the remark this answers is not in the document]
christopherkang: I agree - it's pretty confusing, and I'm wondering if it would be better just as a hidden method (especially because I don't think there is much use for this outside of swapping an array)

christopherkang:
Thanks, @msoeken and @cgranade - I appreciate the help! I've updated the code with the PR'd functions, added examples to the docs. Here's our remaining action items (alongside code review)

- [ ] Naming convention for `ApplySeriesOfOps` and `ApplyOpRepeatedlyOver` (current suggestions from Chris are `ApplyEachToEachSubregister` and `ApplyToEachSubregister`)
- [ ] Decide whether to rename or hide `SwapOrderToPermuteArray`

Let me know if I missed anything!

Standard/src/Arrays/Arrays.qs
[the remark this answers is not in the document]
msoeken: I agree to make this function private for now.

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +283,110 @@ namespace Microsoft.Quantum.Arrays {
...
+        // Check to verify the new ordering actually is a permutation of the indices
+        Fact(_IsPermutation(newOrder), $"The new ordering is not a permutation");
+
+        mutable swaps = new (Int, Int)[0];
+        mutable order = newOrder;
+
+        // for each value, whenever the index and value don't match, swap until it does
+        for (index in 0..Length(order) - 1) {
```
msoeken: I suggest to replace this with `for (index in IndexRange(order)) {`.

Standard/src/Arrays/Arrays.qs
```diff
@@ -280,6 +283,110 @@ namespace Microsoft.Quantum.Arrays {
...
+    ///
+    /// # Output
+    /// The array with the in place swapp applied.
+    ///
+    /// ## Example
+    /// ```qsharp
+    /// // The following returns [0, 3, 2, 1, 4]
+    /// Swapped(1, 3, [0, 1, 2, 3, 4]);
+    function Swapped<'T>(firstIndex: Int, secondIndex: Int, arr: 'T[]) : 'T[] {
+        mutable newArray = arr;
```
msoeken: This line can be removed.

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,256 @@
...
+    /// Nested arrays describing the targets of the op. Each array should contain a list of ints describing 
+    /// the qubits to be used.
+    /// ## register
+    /// Qubit register to be acted upon.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplySeriesOfOps
+    operation ApplyOpRepeatedlyOver(op : (Qubit[] => Unit), targets : Int[][], register : Qubit[]) : Unit 
+    {
+        for (index in 0..Length(targets) - 1)
```
msoeken: Since you don't use `index` separately, you can write `for (target in targets)`

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,256 @@
...
+    /// Nested arrays describing the targets of the op. Each array should contain a list of ints describing 
+    /// the qubits to be used.
+    /// ## register
+    /// Qubit register to be acted upon.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplySeriesOfOps
+    operation ApplyOpRepeatedlyOverA(op : (Qubit[] => Unit is Adj), targets : Int[][], register : Qubit[]) : Unit is Adj
+    {
+        for (index in 0..Length(targets) - 1)
```
msoeken: See above.

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,256 @@
...
+    /// Nested arrays describing the targets of the op. Each array should contain a list of ints describing 
+    /// the qubits to be used.
+    /// ## register
+    /// Qubit register to be acted upon.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplySeriesOfOps
+    operation ApplyOpRepeatedlyOverC(op : (Qubit[] => Unit is Ctl), targets : Int[][], register : Qubit[]) : Unit is Ctl
+    {
+        for (index in 0..Length(targets) - 1)
```
msoeken: See above.

Standard/src/Canon/Combinators/ApplyRepeatedOver.qs
```diff
@@ -0,0 +1,256 @@
...
+    /// Nested arrays describing the targets of the op. Each array should contain a list of ints describing 
+    /// the qubits to be used.
+    /// ## register
+    /// Qubit register to be acted upon.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplySeriesOfOps
+    operation ApplyOpRepeatedlyOverCA(op : (Qubit[] => Unit is Adj+Ctl), targets : Int[][], register : Qubit[]) : Unit is Adj+Ctl
+    {
+        for (index in 0..Length(targets) - 1)
```
msoeken: See above.

Standard/src/Standard.csproj
msoeken: You can leave it to `False` but without committing the change for now.

christopherkang:
Thanks, @msoeken! I appreciate it 😃

I was ideating on potential names - what about ApplyEachToSubarrays (for different ops) and ApplyToSubarrays (for the same op). Open to suggestions!

Standard/src/Standard.csproj
msoeken: Please revert this change in order for us to proceed with the PR.

msoeken:
@christopherkang Thanks for the last changes, all my comments have been addressed. I cannot comment much on the naming of the functions, @cgranade, could you please check whether the names of the new functions are okay.

christopherkang:
Thank you both! @cgranade I renamed the Array function, but wasn't sure what you meant by type conversion function. You're welcome to merge changes, once checks pass! I can also make those edits, I'll just need some clarification. Whatever's easier for you! 😄

cgranade:
My apologies for missing the GitHub notification about your comment, @christopherkang! Merging in now, thank you so much for your contribution. 💕

christopherkang:
No worries @cgranade - are you aware of any areas that could be refactored with these operations? If so, I'll start working on a new PR

cgranade:
I think your most recent round addressed everything, @christopherkang, great to have your contribution on board for the next release! My apologies again for missing the notification.

christopherkang:
Awesome! Also, @cgranade, I mean do you know of any code in the wider Q# codebase that could make use of this operation?

PR #163, opened 2019-10-13
Updating to .NET Core 3.0

bettinaheim:


Chemistry/src/Jupyter/Jupyter.csproj
```diff
@@ -26,8 +26,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Jupyter.Core" Version="1.1.12077" />
-    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.9.1909.3002" />
-    <PackageReference Include="Microsoft.Quantum.IQSharp.Core" Version="0.9.1909.3002" />
+    <PackageReference Include="Microsoft.Quantum.Simulators" Version="0.9.1910.1107-alpha" />
+    <PackageReference Include="Microsoft.Quantum.IQSharp.Core" Version="0.8.1907.1701" />
```
cgranade: Is this the right version for IQ# Core?

cgranade:
Quick thought, @bettinaheim, but we encountered some migration issues with MS.Jupyter.Core and IQ# such that we had to update those packages; would it make sense to hold this PR off until microsoft/iqsharp#43 is in, so that we can update Microsoft.Quantum.Chemistry.Jupyter?

Chemistry/src/Jupyter/Jupyter.csproj
bettinaheim: We need this alpha package before we can update IQ#, but the 0.8 vs 0.9 was an accidental effect on merging. Fixed that.

PR #164, opened 2019-10-14
Enabling all source inclusion with source link

vadym-kl:
This PR enables storing compressed source files inside `PDB` files, so they can later be accessed using `System.Reflection.Metadata` without the need for internet connection.


vadym-kl:
All sources are indeed packaged into `pdb`, as shown by `dotPeek`:
image

PR #179, opened 2019-11-21
Simplify classification API.

cgranade:


MachineLearning/src/Runtime/Circuits.qs
```diff
@@ -342,165 +201,34 @@ namespace Microsoft.Quantum.MachineLearning {
...
-		{
-			let dL =Microsoft.Quantum.Math.Ceiling(Lg(IntAsDouble (Length(getData(Head(samples))))));
-			if (N < dL)
-			{
-				set N = dL;
-			}
-		}
-		set ret = new (Double,Int)[sC];
-		mutable ir = 0;
+	/// TODO
```
bettinaheim: Looks like there was something here you still wanted to do?

MachineLearning/src/Runtime/Types.qs
```diff
@@ -45,6 +45,25 @@ namespace Microsoft.Quantum.MachineLearning {
...
+	/// # Input
+	/// ## schedule
+	/// A sampling schedule whose length is to be returned.
+	///
+	/// # Output
+	/// The number of elements in the given sampling schedule.
+	function ScheduleLength(schedule : SamplingSchedule) : Int {
+		mutable length = 0;
+		for (range in schedule!) {
+			for (index in range) {
```
bettinaheim: I think we might want to have something like a Length function for ranges.

MachineLearning/src/Runtime/Training.qs
```diff
@@ -6,6 +6,101 @@ namespace Microsoft.Quantum.MachineLearning {
     open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Canon;
 
+	/// # Summary
+	/// Semi-greedily find a bias value that leads to near-minimum misclassification score.
```
bettinaheim: I have no idea what semi-greedily means.

MachineLearning/src/Runtime/Training.qs
```diff
@@ -6,6 +6,101 @@ namespace Microsoft.Quantum.MachineLearning {
...
+				set ipro += 1;
+			}
+		}
+		// Exit early if we can find a perfect classification.
+		if (max0 <= min1) {
+			return 0.5 * (1.0 - max0 - min1);
+		}
+		mutable mBest = Length(probabilities);
+		mutable bBest = bias;
+		mutable bLeft = 0.5 - max0;
```
bettinaheim: I don't know where our style guide lands; do we always want to declare only one variable at a time? Because at least bLeft and bRight sound like they belong together, and I would suggest declaring them in one statement.

MachineLearning/src/Runtime/Classification.qs
```diff
@@ -1,35 +1,114 @@
...
+	/// resulting from estimating classification frequencies.
+	///
+	/// # Output
+	/// The label inferred from each classification probability.
+	function InferredLabels(bias : Double, probabilities : Double[]): Int[] {
+		return Mapped(InferredLabel(bias, _), probabilities);
+	}
+
+	/// # Summary
+	/// Get a list of all the classification probabilities. In the from of (prob1,label) pairs. THIS operation is IN DEPRECATION
```
bettinaheim: The Capitalization of "THIS" seems a bit arbitrary to me.

MachineLearning/src/Runtime/Types.qs
cgranade: That would be really nice to have, definitely.

MachineLearning/src/Runtime/Training.qs
cgranade: This doc comment was migrated from elsewhere, but I agree needs further iteration. Thanks for flagging that.

MachineLearning/src/Runtime/Training.qs
cgranade: Agreed, I would like to refactor this function in particular much further.

MachineLearning/src/Runtime/Classification.qs
cgranade: Yes, the whole comment about deprecation should be removed in fact.

MachineLearning/src/Runtime/Training.qs
cgranade: If it's OK, I'd like to do so in a follow-up PR.

MachineLearning/src/Runtime/Circuits.qs
cgranade: Yes, I fully intend on further refactoring this operation, as it's somewhat redundant with other operations in the library.

PR #184, opened 2019-12-04
Consolidate exact and approximate multiplexing logic.

cgranade:
This PR consolidates the new approximate multiplexing and state preparation logic introduced with the QML feature into the existing functionality provided by the standard library. As this is going into the QML feature branch (see #176), unit testing of the new approximate multiplexing logic has been temporarily deferred to the broader unit testing task in that draft PR.


cgranade:
Thanks for looking over! I didn't change anything about `ReflectAboutInteger` in this PR, so it probably just didn't show up in the changes for this PR. It's definitely still there, though. In any case, thank you again for the review!

PR #185, opened 2019-12-07
Add placeholder for QML unit tests.

cgranade:
This PR adds a unit test to the QML library in preparation for testing QML functionality more broadly. This PR also updates the QML library and its dependencies to the new 0.10.1912.501 version of the compiler and runtime, in order to use the new `@Test("…")` attribute.


alexeib2:
Looks good to me.

PR #187, opened 2019-12-12
Use new UDTs to consolidate and simplify training / validation API.

cgranade:
This PR introduces new UDTs to help consolidate and simplify the training and validation API, and makes changes that API to be more internally consistent (e.g.: choosing a uniform ordering for inputs). To enable these changes, this PR also introduces two new array functions into the standard library, `Chunks` and `Where`.


MachineLearning/src/DataModel/Interop.cs
```diff
@@ -176,7 +176,7 @@ public static List<long[]> PartialLocalLayer(long[] indices, char pauli)
 
         /// <summary>
         /// Creates a cyclic block of nQubits controlled rotations that starts
-        /// with contol qubit (nQubits-1), target qubit (cspan-1) % n , followed by the 
+        /// with contol qubit (nQubits-1), target qubit (cspan-1) % n , followed by the
         /// ladder of entanglers with control qubit iq and target qubit (iq+cspan) % n
```
martinquantum: /// with control qubit (nQubits-1), target qubit (cspan-1) % nQubits , followed by a
/// ladder of entanglers with control qubits iq and target qubit (iq+cspan) % nQubits

MachineLearning/src/Runtime/Examples.qs
```diff
@@ -1,96 +1,94 @@
...
-				1)), LabeledSample(([0.533848, 0.501219, 0.4732, 0.489762], 
-				0)), LabeledSample(([0.530036, 0.577194, 0.452731, 0.425375], 
-				1)), LabeledSample(([0.595573, 0.439349, 0.494919, 0.455325], 
-				1)), LabeledSample(([0.584424, 0.557699, 0.438769, 0.393576], 
-				1)), LabeledSample(([0.544759, 0.441244, 0.494108, 0.514196], 
-				0)), LabeledSample(([0.552072, 0.545641, 0.487013, 0.400388], 1))				
-		];
-		return LabeledSampleContainer(ret);
-	}
+    operation IrisTrainingData() : LabeledSample[] {
```
martinquantum: I'm wondering if it would be better to separate code and data and to store the data into a separate file?

cgranade: Agreed, though there is some precedent for classical ML libraries making common testing datasets available as functions to help users write their own benchmarks. For instance, scikit-learn provides the [`load_iris` function](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_iris.html). Perhaps it might make sense to keep this but under a new subnamespace (i.e.: `Microsoft.Quantum.MachineLearning.Datasets`)?

martinquantum: As per discussion in the Libraries meeting, this is a Q# convenience function to provide access to small data sets (like IRIS, half moon etc), so we will leave this as is

martinquantum:
merging in this PR. Thanks for refactoring, Chris!

PR #190, opened 2019-12-16
Minor doc fixes

numpde:


Standard/src/Arithmetic/Modular.qs
```diff
@@ -27,12 +27,14 @@ namespace Microsoft.Quantum.Arithmetic {
     /// Integer y in `LittleEndian` format that `increment` a is added to.
     ///
     /// # See Also
-    /// - IncrementPhaseByModularInteger
+    /// - Microsoft.Quantum.Arithmetic.IncrementPhaseByModularInteger
```
cgranade: This is interesting, as the documentation generation tool should correctly resolve the namespace here; I've filed as a separate bug at microsoft/qsharp-compiler#267, but I like keeping your workaround for now. Thanks!

Standard/src/Arithmetic/Modular.qs
```diff
@@ -50,13 +52,13 @@ namespace Microsoft.Quantum.Arithmetic {
...
-    /// Let us denote `increment` by a, `modulus` by N and integer encoded in `target` by y
+    /// Let us denote `increment` by a, `modulus` by N and integer encoded in `target` by y.
     /// Then the operation performs the following transformation:
     /// |y⟩ ↦ |y+a (mod N)⟩
-    /// Integers are encoded in little-endian format in QFT basis
+    /// Integers are encoded in little-endian format in QFT basis.
     ///
     /// # See Also
-    /// - Microsoft.Quantum.Canon.ModularIncrementLE
+    /// - Microsoft.Quantum.Canon.ModularIncrementLE (deprecated)
```
cgranade: The `# See Also` block doesn't support additional information such as deprecation notices, unfortunately. In this case, perhaps it would be better to change the See Also link instead?
```suggestion
    /// - Microsoft.Quantum.Arithmetic.ModularIncrementByInteger
```

Standard/src/Arithmetic/Modular.qs
```diff
@@ -153,11 +155,11 @@ namespace Microsoft.Quantum.Arithmetic {
     }
 
     /// # Summary
-    /// The same as ModularAddProductLE, but assumes that summand encodes
+    /// The same as ModularAddProductLE (deprecated), but assumes that summand encodes
     /// integers in QFT basis
     ///
     /// # See Also
-    /// - Microsoft.Quantum.Canon.ModularAddProductLE
+    /// - Microsoft.Quantum.Canon.ModularAddProductLE (deprecated)
```
cgranade:
```suggestion
    /// - Microsoft.Quantum.Arithmetic.MultiplyAndAddByModularInteger
```

Standard/src/Arithmetic/Modular.qs
```diff
@@ -153,11 +155,11 @@ namespace Microsoft.Quantum.Arithmetic {
     }
 
     /// # Summary
-    /// The same as ModularAddProductLE, but assumes that summand encodes
+    /// The same as ModularAddProductLE (deprecated), but assumes that summand encodes
```
cgranade: Similarly here, it might help to change the reference to point to the new name of the `ModularAddProductLE` operation:
```suggestion
    /// The same as MultiplyAndAddByModularInteger, but assumes that summand encodes
```

Standard/src/Arithmetic/Modular.qs
numpde: I admit this was rather meant to point out the issue. IMO, it's better to explicit the namespace if there is "danger" that it might change (again), but it doesn't seem to be the case here.

cgranade: Fair enough! The issue is now being tracked in any case, thanks again for pointing it out.

Standard/src/Arithmetic/Modular.qs
numpde: Maybe it is still worthwhile to reference the deprecated implementation in the `# Remarks` block. 

Then, `# See Also` of [IncrementPhaseByModularInteger](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arithmetic.incrementphasebymodularinteger) should reference [IncrementByModularInteger](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arithmetic.incrementbymodularinteger), and vice versa. 

(Being new to this, I don't know how to take it from here; do I mark this as "resolved" and initiate a new PR?)

cgranade: With the new `@Deprecated()` attribute, we've definitely improved on our ability to track deprecation and offer better warnings to users. Given that, I would suggest that `/// # See Also` blocks only refer to non-deprecated operations and functions, so as to present a consistent view of how the libraries evolve going forward.

With respect to moving forward with your contribution, there's no need to make a new PR, as this PR will update as you make changes to your `numpde:patch-2` branch. For instance, if you use the "Commit suggestion" button above, that will create a new commit on your branch co-authored by the two of us, updating your PR accordingly. We have a few more details up in the [contribution guide](https://docs.microsoft.com/quantum/contributing/pull-requests), but I'm also happy to answer anything that I can.

numpde: Is there a preferred way, `<xref:uid_of_the_topic>` vs `@"uid_of_the_topic"`?

numpde: I've gone over it [again](https://github.com/microsoft/QuantumLibraries/pull/190/commits/37266fec66bfab44402f1acd4a49bd77ca921c91).

numpde:
Thanks for your comments; another iteration is in.

PR #196, opened 2019-12-26
New Hadamard and SWAP test operations.

cgranade:
This PR adds new functionality to Microsoft.Quantum.Characterization for the Hadamard and SWAP tests. Each of these new operations is compatible with EstimateFrequencyA, and hence can be accelerated using emulation functionality where available.

Currently a work in progress, opening PR for initial feedback. Still to do:
- [x] unit tests
- [x] finish API documentation comments


Standard/src/Characterization/Distinguishability.qs
```diff
@@ -0,0 +1,234 @@
...
+    /// $U$ is the unitary representation of the action of `preparation1`,
+    /// and where $V$ corresponds to `preparation2`.
+    ///
+    /// # References
+    /// - Aharonov *et al.* [quant-ph/0511096](https://arxiv.org/abs/quant-ph/0511096).
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Characterization.EstimateRealOverlapBetweenStates
+    /// - Microsoft.Quantum.Characterization.EstimateOverlapBetweenStates
+    operation EstimateImagOverlapBetweenStates(
```
martinquantum: Imag or Imaginary? Do we have a precedent for this elsewhere in the API docs?

cgranade: Good point, thanks for the catch. I guess I was thinking of the [usage in `Microsoft.Quantum.Math.Complex`](https://github.com/microsoft/qsharp-runtime/blob/a91439ff65bd93e790d81cef73b6abb3f1589e22/src/Simulation/Intrinsic/Math/Types.qs#L14), which has a `Real` and an `Imag` named item.

PR #200, opened 2020-01-04
Simplify gradient estimation and classifier structure generation

cgranade:
This PR includes recent changes from master into #176 and uses these changes to simplify gradient estimation. This PR also moves classifier structure specification from C# methods to Q# functions, making them easier to use from other .NET languages or from Python.

I will shortly make a new PR on the samples repo as well with the needed matching changes to the samples.


cgranade:
As an update, moving the circuit structure generation logic out of the C# library into Q# means that the C# part is no longer actively used, such that I've gone on and consolidated everything back into a single-project layout for the QML library.

cgranade:
Thanks for the review, @alexeib2!

> Good refactoring. The function name _EstimateFiniteDifference does not strike me as the best choice.
> 
> 1. why 'Finite'?
> 2. and besides, 'FiniteDifference' has … ahem... Different connotation in calculus and engineering in general.

This is a private operation, but all the same, good to rename to be consistent; I'll go on and do that.

> I was wondering about the practice of creating empty arrays and then populating them by iterative concatenation.
> For example
> set complexCoefficients += [ComplexPolar(coef, ang)];
> where the array complexCoefficients is inialized as empty. A few more instances like that across the code.
> Doesn't this create a steady stream of temporaries that then need to be garbage-collected?

Thankfully, this isn't the case. Both approaches are indeed 𝑂(𝑛), as can be confirmed using BenchmarkDotNet:

image

> What is the wisdom of doing this in a situation when the size of the array is known upfront?

My motivation here has been to prioritize maintainability and readability in implementations, but I'm also happy to switch to using `Enumerated` and `w/=` given the overhead in this case (a small coefficient, but still can be relevant).

---

(As a quick tip, it makes it much easier to find and respond to comments if you attach them to particular lines in the source code. Thanks!)

cgranade:
@alexib2: Thanks again for your review. I've addressed your feedback in the most recent commit, if you could please review and mark as approved then we can get this in. Thanks!

MachineLearning/src/Convert.qs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.MachineLearning {
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Math;
+
+    function unFlattenSchedule(sc : Int[][]) : SamplingSchedule
```
bettinaheim: What's the reason for the lower case here? I thought global declarations should usually be upper case? Same everywhere below.

cgranade: That's the subject of the next PR!

PR #204, opened 2020-01-13
Add new MeasureIfAllZeros operation.

cgranade:


cgranade:
@tcNickolas: I've gone on and added this operation as per your feedback, please let me know if it addresses your usecase before I merge it in. Thanks!

Standard/src/Measurement/Registers.qs
```diff
@@ -72,4 +72,27 @@ namespace Microsoft.Quantum.Measurement {
         return ForEach(M, targets);
     }
 
+    /// # Summary
+    /// Measures a register of qubits is in the all-zeros state.
```
tcNickolas: I'm not sure how to parse this sentence... 
```suggestion
    /// Measures a register of qubits and returns true if it is in the all-zeros state or false otherwise.
```

Standard/src/Measurement/Registers.qs
```diff
@@ -72,4 +72,27 @@ namespace Microsoft.Quantum.Measurement {
...
+    /// # Input
+    /// ## register
+    /// The register of qubits to be measured.
+    ///
+    /// # Output
+    /// `true` if and only if the register is measured to be in the
+    /// $\ket{00 \cdots 0}$ state.
+    ///
+    /// # Remarks
+    /// This operation does not reset its qubits, but projects them to a
```
tcNickolas: This behavior differs from that of `MeasureInteger` - any specific reason why? It seems more neat to be able to write "return MeasureIfAllZeros(register)" than to measure into a variable, then call ResetAll, then return that variable.

Standard/src/Measurement/Registers.qs
```diff
@@ -72,4 +72,27 @@ namespace Microsoft.Quantum.Measurement {
         return ForEach(M, targets);
     }
 
+    /// # Summary
+    /// Measures a register of qubits is in the all-zeros state.
+    ///
+    /// # Description
+    /// Given a register of qubits, measures if that register is in the state
```
tcNickolas:
```suggestion
    /// Given a register of qubits, checks if that register is in the state
```
?

Standard/src/Measurement/Registers.qs
cgranade: My intent was to match the semantics of other operations in the Microsoft.Quantum.Measurement namespace; the operations in Microsoft.Quantum.Arithmetic don't follow the same convention (they should be reconciled, but that's a slightly separate discussion).

For the usecase of returning after, I could imagine something like `ResetAfter<'T>(op : (Qubit[] => 'T), target : Qubit[]) : 'T { let returnValue = op(target); ResetAll(target); return returnValue; }` such that you could do `return ResetAfter(MeasureIfAllZeros(register));`. That would be less efficient in terms of the number of measurements than `return All(ForEach(MResetZ, target))`, though.

tcNickolas: In the last code snippet, did you mean something like `All(IsResultZero, ForEach(MResetZ, target))`?

From the end user point of view, it might be still more convenient to stick with MeasureInteger than to use MeasureIfAllZeros with variable assignment or to chain together several operations from different namespaces...

cgranade: Yes, thank you; that's the danger of writing code quickly on the way to a meeting, I suppose. Anyway, I think there's two distinct issues that are getting a bit conflated here:

- Existing measurement operations expose different reset semantics based on whether they are in Microsoft.Quantum.Measurement or Microsoft.Quantum.Arithmetic.
- Both kinds of reset semantics have independent usecases, and should be made available to uesrs.

msoeken:
@cgranade, could you please submit an issue for an API review of the proposed operation. We can then schedule it for API review in order to complete this PR. If the PR is no longer of interest, we can close it alternatively. Thanks

cgranade:
> @cgranade, could you please submit an issue for an API review of the proposed operation. We can then schedule it for API review in order to complete this PR. If the PR is no longer of interest, we can close it alternatively. Thanks

Sounds good, can do that. I think this also originally came your suggestion, @tcNickolas, if you'd prefer to file the issue?

tcNickolas:
@cgranade I'll admit I have no recollection of suggesting this... My best guess it had something to do with checking whether the oracles modify their input registers, this could've been something I did two years ago?

I think it will be better if you do the proposal - from the comments it looks like you had thoughts on offering different variants of this operation rather than just one? And then I chime in on the API review step

PR #205, opened 2020-01-13
Final prep for QML review

cgranade:
This PR refactors a bit more, removing unused functions and operations, and swapping input orderings to be more consistent. Following this PR, I believe that the QML library will be ready for initial API review, after which we can proceed to improve API documentation and unit testing coverage.


cgranade:
Thanks for the review!

PR #207, opened 2020-01-16
Fix #206 and improve Trotter–Suzuki code quality.

### LINKED ISSUE #206: API documentation for DecomposeIntoTimeStepsCA is confusing.
As per discussions in #178, the notation used by `DecomposeIntoTimeStepsCA` is different than the notation used in the reference cited by that operation (namely, [quant-ph/0508139](https://arxiv.org/pdf/quant-ph/0508139.pdf)).

> Thanks for pointing out the difference between the paper quant-ph/0508139 and the proposed change Chris!
> 
> We can better match the notation of the paper by putting the missing factor of two into
> `let stepSizeOuter = _TrotterStepSize(order/2);`
> &
> `return 1.0 / (4.0 - PowD(4.0, 1.0 / (IntAsDouble(2*order) - 1.0)));`
> 
> instead of the proposed change
> `let stepSizeOuter = _TrotterStepSize(order);`
> &
> `return 1.0 / (4.0 - PowD(4.0, 1.0 / (IntAsDouble(order) - 1.0)));`
> 
> That said, proposed change is indeed consistent with the paper.
> —@guanghaolow

This creates confusion in trying to specify inputs to `DecomposeIntoTimeStepsCA`, such that a remark should be added to clarify this distinction.

cgranade:
This PR fixes #206 by adding in an appropriate `/// # Remark` noting the 𝑘–2𝑘 dichotomy. This PR also improves code quality in the Trotter–Suzuki decomposition feature more generally and renames `DecomposeIntoTimeStepsCA` → `DecomposedIntoTimeStepsCA` to reflect that this feature is a exposed as a **function** that returns a new evolution operation.

### DIFF
--- a/Standard/src/Canon/Enumeration/Trotter.qs
+++ b/Standard/src/Canon/Enumeration/Trotter.qs
@@ -31,13 +31,12 @@ namespace Microsoft.Quantum.Canon {
     /// op(0, deltaT, target);
     /// op(1, deltaT, target);
     /// ```
-	/// and
-	/// ```qsharp
+    /// and
+    /// ```qsharp
     /// _Trotter1ImplCA((2, op), deltaT, target);
     /// ```
     operation _Trotter1ImplCA<'T> ((nSteps : Int, op : ((Int, Double, 'T) => Unit is Adj + Ctl)), stepSize : Double, target : 'T) : Unit is Adj + Ctl {
-        for (idx in 0 .. nSteps - 1)
-        {
+        for (idx in 0 .. nSteps - 1) {
             op(idx, stepSize, target);
         }
     }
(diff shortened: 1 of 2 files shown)


Standard/src/Canon/Enumeration/Trotter.qs
```diff
@@ -157,24 +162,27 @@ namespace Microsoft.Quantum.Canon {
     /// the first parameter `Double` is the integration step size, and the
     /// second parameter is the target acted upon.
     ///
+    /// # Remarks
+    /// This operation uses a different indexing convention than that of
+    /// [quant-ph/0508139](https://arxiv.org/abs/quant-ph/0508139). In
+    /// particular, `DecomposedIntoTimeStepsCA(_, 4)` corresponds to the
+    /// propagator $S_2(\lambda)$ in quant-ph/0508139.
+    ///
```
guanghaolow: `DecomposedIntoTimeStepsCA(_, 4)` actually corresponds to $S_4(\lambda)$ in quant-ph/0508139. 

The difference in indexing occurs in the definition of _TrotterStepSize(order: Int). 

Whereas we set `let stepSizeOuter = _TrotterStepSize(order);` in the implementation of `_TrotterArbitraryImplCA<'T>`, they define the function _AlternateStepSize(order: Int) such that our `stepSizeOuter` is equal to `_AlternateStepSize(order/2)`.

cgranade: Thanks, my apologies for misunderstanding where the index convention differed. I'll fix that, then.

cgranade:
@guanghaolow: Would you be willing to take a look at this again? Thanks for the help!

PR #215, opened 2020-01-27
Improve API documentation for ControlledOnBitString

tcNickolas:
This change adds description, remarks and examples sections.

### DIFF
--- a/Standard/src/Canon/Utils/ControlledOnBitString.qs
+++ b/Standard/src/Canon/Utils/ControlledOnBitString.qs
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
     /// # Summary
     /// Returns a unitary operator that applies an oracle on the target register if the control register state corresponds to a specified bit mask.
     ///
+    /// # Description
+    /// Given a Boolean array `bits` and a unitary operation `oracle`, the return of this function will perform the following steps:
+    /// * apply an X gate to each qubit of the control register that corresponds to `false` element of the `bits`;
+    /// * apply `Controlled oracle` to the control and target registers;
+    /// * apply an X gate to each qubit of the control register that corresponds to `false` element of the `bits` again to return the control register to the original state.
+    ///
     /// # Input
     /// ## bits
     /// Boolean array.
     /// ## oracle
-    /// Unitary operator.
+    /// Unitary operator to be applied on the target register.
     ///
     /// # Output
     /// A unitary operator that applies `oracle` on the target register if the control register state corresponds to the bit mask `bits`.
+    ///
+    /// # Remarks
+    /// The length of `bits` and `controlRegister` must be equal.
+    /// 
+    /// The following are equivalent:
+    /// ```qsharp
+    /// (ControlledOnBitString(bits, oracle))(controlRegister, targetRegister);
+    /// ```
+    /// and
+    /// ```qsharp
+    /// within {
+    ///     ApplyPauliFromBitString(PauliX, false, bits, controlRegister);
+    /// } apply {
+    ///     Controlled oracle(controlRegister, targetRegister);
+    /// }
+    /// ```
+    /// 
+    /// # Example
+    /// Prepare a state $\frac{1}{2}(\ket{00} - \ket{01} + \ket{10} + \ket{11})$:
+    /// ```qsharp
+    /// using (register = Qubit[2]) {
+    ///     ApplyToEach(H, register);
+    ///     (ControlledOnBitString([false], Z))(register[0..0], register[1]);
+    /// }
+    /// ```
     function ControlledOnBitString<'T> (bits : Bool[], oracle : ('T => Unit is Adj + Ctl)) : ((Qubit[], 'T) => Unit is Adj + Ctl)
     {
         return ControlledOnBitStringImpl(bits, oracle, _, _);
--- a/Standard/src/Canon/Utils/ControlledOnBitString.qs
+++ b/Standard/src/Canon/Utils/ControlledOnBitString.qs
@@ -38,11 +38,15 @@ namespace Microsoft.Quantum.Canon {
     /// Returns a unitary operator that applies an oracle on the target register if the control register state corresponds to a specified bit mask.
     ///
     /// # Description
-    /// Given a Boolean array `bits` and a unitary operation `oracle`, the output of this function
-    /// is an operation that performs the following steps:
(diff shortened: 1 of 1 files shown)


tcNickolas:
I did a (hopefully) matching manual change to the docs site to check that the formatting renders the way I'd like it to - you can preview it at https://review.docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.canon.controlledonbitstring?view=qsharp-preview&branch=tcNickolas-patch-1.

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
...
     ///
+    /// # Description
+    /// Given a Boolean array `bits` and a unitary operation `oracle`, the return of this function will perform the following steps:
+    /// * apply an X gate to each qubit of the control register that corresponds to `false` element of the `bits`;
+    /// * apply `Controlled oracle` to the control and target registers;
+    /// * apply an X gate to each qubit of the control register that corresponds to `false` element of the `bits` again to return the control register to the original state.
+    ///
     /// # Input
     /// ## bits
     /// Boolean array.
```
cgranade: I recognize that this was not part of your PR, but the description of the `bits` input as just "boolean array" feels a bit out of place with all the other improvements that you made. Perhaps we can improve it as well?
```suggestion
    /// The bitstring to control the given quantum operation on.
```

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
     /// # Summary
     /// Returns a unitary operator that applies an oracle on the target register if the control register state corresponds to a specified bit mask.
     ///
+    /// # Description
+    /// Given a Boolean array `bits` and a unitary operation `oracle`, the return of this function will perform the following steps:
```
cgranade: I'd suggest changing "return" to "output" here to be consistent with what the different section headers are called, but I'll readily admit that's pretty minor.
```suggestion
    /// Given a Boolean array `bits` and a unitary operation `oracle`, the output of this function is an
    /// operation that performs the following steps:
```

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
...
+    /// * apply an X gate to each qubit of the control register that corresponds to `false` element of the `bits`;
```
cgranade:
```suggestion
    /// * apply an `X` operation to each qubit of the control register that corresponds to `false` element of the `bits`;
```

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
...
+    /// * apply `Controlled oracle` to the control and target registers;
+    /// * apply an X gate to each qubit of the control register that corresponds to `false` element of the `bits` again to return the control register to the original state.
```
cgranade:
```suggestion
    /// * apply an `X` operation to each qubit of the control register that corresponds to `false` element of the `bits` again to return the control register to the original state.
```

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
...
-    /// Unitary operator.
+    /// Unitary operator to be applied on the target register.
     ///
     /// # Output
     /// A unitary operator that applies `oracle` on the target register if the control register state corresponds to the bit mask `bits`.
+    ///
+    /// # Remarks
+    /// The length of `bits` and `controlRegister` must be equal.
+    /// 
+    /// The following are equivalent:
```
cgranade:
```suggestion
    ///
    /// # Example
    /// The following are equivalent:
```

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -37,14 +37,45 @@ namespace Microsoft.Quantum.Canon {
...
+    /// ```qsharp
```
cgranade: I've gotten conflicting messages on whether `Q#` or `qsharp` here is correct for our docs pipeline; it may be good to double-check just in case.

tcNickolas: I checked that ```qsharp generates a snippet with language marked as Q# on the site, and we have other places where it is used, so I figured it's the right way to do it. Haven't tried Q#, though.

tcNickolas:
For `# Description` vs `# Remarks`, I opted for description, since that puts this information above the information on parameter types. I've observed that this info is really important for the reader to understand *what* is going on, not just *how* it is done - the idea of controlling on a different bit string can be unfamiliar.

cgranade:
> For `# Description` vs `# Remarks`, I opted for description, since that puts this information above the information on parameter types. I've observed that this info is really important for the reader to understand _what_ is going on, not just _how_ it is done - the idea of controlling on a different bit string can be unfamiliar.

I think we can achieve that goal — namely, of explaining the concept of controlling on a bitstring to those that are unfamiliar — a bit more directly, perhaps?

```
/// The operation output by this function can be
/// simulated by a unitary transformation $U$ such that
/// $$
/// \begin{align}
///     U \ket{b_0 b_1 \cdots b_{n - 1}} \ket{\psi} = \ket{b_0 b_1 \cdots b_{n-1}} \otimes \begin{cases}
///         V \ket{\psi} & (b_0 b_1 \cdots b_{n - 1}) = \texttt{bits} \\\\
///         \ket{\psi} & \textrm{otherwise}
///     \end{cases},
/// \end{align}
/// $$
/// where $V$ is a unitary matrix simulating the action of the `oracle`
/// operation.
/// By contrast, output the `Controlled` functor represents the special case where
/// `bits` is equal to `[true, true, ..., true]`.
```

A link back to the new manual section that @geduardo is working on may be helpful to tie this back to the more general concept of controlling operations as well, so that the reference documentation isn't stuck with needing to do all of the conceptual heavy lifting as well.

tcNickolas:
Good point, this will also allow us to show how to do formulas in the description. I've updated the PR and the preview branch accordingly.

The link back will have to wait until the new section is in...

PR #222, opened 2020-02-03
Fixed QsharpDocsOutputPath property in build.ps1.

cgranade:


cgranade:
Thanks for taking a look @tcNickolas. Once this gets in, can use to update the docs for microsoftdocs/quantum-docs-pr#683 as per discussions there.

PR #233, opened 2020-02-21
Addressing feedback from @bettinaheim.

cgranade:
This PR addresses feedback from @bettinaheim on #176.


cgranade:
Merging in as admin so as to incorporate feedback into #176.

PR #234, opened 2020-02-22
Add verbose option to TrainingOptions UDT, cache state preparation

cgranade:
This PR uses https://github.com/microsoft/qsharp-runtime/pull/151 to add a new verbosity option to the `TrainingOptions` UDT. This option allowed for diagnosing https://github.com/microsoft/Quantum/issues/315 more precisely, during which it was identified that the caching of input encoders could improve performance. This PR also implements that solution.


MachineLearning/src/Training.qs
```diff
@@ -93,7 +93,7 @@ namespace Microsoft.Quantum.MachineLearning {
         let labels = Mapped(_Label, samples);
 
         for ((idxModel, model) in Enumerated(models)) {
-            Message($"Beginning training at start point #{idxModel}...");
+            options::VerboseMessage($"  Beginning training at start point #{idxModel}...");
```
bettinaheim: Is the whitespace intentional?

MachineLearning/src/Types.qs
```diff
@@ -204,9 +207,14 @@ namespace Microsoft.Quantum.MachineLearning {
...
         MaxEpochs: Int,
         MaxStalls: Int,
-        StochasticRescaleFactor: Double
+        StochasticRescaleFactor: Double,
+        VerboseMessage: (String -> Unit)
     );
 
+    // Used only to populate default training options to ignore verbose
+    // messages.
+    function _Ignored(value : String) : Unit {}
```
bettinaheim: Don't we have a type parametrized ignore function in the standard libraries?

MachineLearning/src/Training.qs
cgranade: It is, yes. I was using that to indent subtasks so that it's easier to see the structure of a training run at the console.

MachineLearning/src/Types.qs
cgranade: We do, yes. I was thinking it was an operation for some reason. I guess I did need more caffeine after all. Thanks for catching that!

cgranade:
> Looks good, aside from the package numbers.

Thanks for the review! The package numbers should work, since that's the beta package built from https://github.com/microsoft/qsharp-runtime/pull/151, and since the end-to-end build will clobber package versions anyway.

PR #237, opened 2020-03-09
Add devcontainer config.

cgranade:
This PR adds a .devcontainer configuration that can be used with either VS Code Remote Development or Visual Studio Online, making it easier to have reproducible builds for development and contributions. In particular, as per discussions with @KittyYeungQ, this will make it easier for first-time contributors to get up and running quickly (especially at events), without having to worry about local setups.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,11 @@
+{
+    "image": "mcr.microsoft.com/quantum/iqsharp-base:latest",
+    "containerEnv": {
+        "IQSHARP_HOSTING_ENVIRONMENT": "libraries-devcontainer"
+    },
+    "extensions": [
+        "ms-dotnettools.csharp",
+        "quantum.quantum-devkit-vscode",
+        "ms-python.python"
+    ]
+}


cgranade:
> Is this affected by [microsoft/qsharp-compiler#323](https://github.com/microsoft/qsharp-compiler/issues/323)?
> 
> Possibly a silly question: This using the IQ# docker image, is there a way to run a Jupyter notebook on the libraries from the dev container?

Not silly at all! I forget where we have it in the docs, but you need the additional argument `--ip 0.0.0.0` when running Jupyter Notebooks from inside a Docker image, such as when using the samples from VS Code Remote Development or VS Online. Concretely, the steps go something like this:

- Run `jupyter notebook --ip 0.0.0.0`
- If using VS Code Remote Development, ensure the port is forwarded from the activity bar (typically on the left-hand side)
- If using VS Online, we're currently blocked by https://github.com/MicrosoftDocs/vsonline/issues/347.
- Copy the token from the terminal, and go to the port forwarded by remote development. Paste the token into the Jupyter login window.

I apologize that these steps are written in a little bit of a hurry, but I'm happy to expand on later or answer any questions. Thanks!

PR #246, opened 2020-05-11
Convert _ functions and operations to use new access modifiers, remove 6-month deprecated callables. (#240)

cgranade:
This PR merges the feature work at #240 onto master.

PR #248, opened 2020-05-12
Remove Reset.qs, move to qsharp-runtime.

cgranade:
This PR removes Reset.qs from the libraries in preparation for moving this file to qsharp-runtime. Note that this PR should not be merged in before #246.


Standard/src/Diagnostics/Facts.qs
```diff
@@ -22,10 +22,36 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    /// # See Also
+    /// - Microsoft.Quantum.Diagnostics.Fact
+    ///
+    /// # Example
+    /// The following Q# code will print "Hello, world":
+    /// ```Q#
+    /// Contradiction(2 == 3, "2 is not equal to 3.");
+    /// Message("Hello, world.");
+    /// ```
+    function Contradiction(actual : Bool, message : String) : Unit {
```
swernli, line 51: Nice! I was hoping this would get added.

cgranade:
Looks like the build is failing because https://github.com/microsoft/qsharp-runtime/pull/199 needs to be completed first. Let me merge that one since you approved, @swernli, and then I'll update this PR to use the CI version built from that PR. Thanks!

PR #256, opened 2020-05-31
Fix typos in API documentation and code comments

tcNickolas:


tcNickolas:
@cgranade Could you please take a look? This seems like it should be quite trivial to review, even though it touches multiple files...

cgranade:
Thanks for the ping, @tcNickolas; all too easy for notifications to get lost. Reviewed and it looks great. Thanks!

PR #260, opened 2020-06-04
Initial version of chemistry format conversion tool.

cgranade:
This PR begins making a command-line conversion tool for different electronic structure problem serialization formats (e.g.: Broombridge, Liquid, and FCIDUMP). In support of that feature, this PR introduces a number of breaking changes to chemistry serialization and deserialization logic to abstract away the format used to represent electronic structure problems from their on-disk representation, making it easier for us to revise and improve Broombridge between versions of the spec, and for legacy formats like FCIDUMP to be supported alongside. As a side effect, this PR also works to simplify index conversion logic to make it easier to support other indexing conventions directly within Broombridge in future versions.


dgwvfxn:
I see the coulomb_respulsion and energy_offset values are assigned incorrectly, (they mix-matched), could you fix that? These values are not as important as the integrals, but a lot of chemistry solvers still use coulomb_respulsion to assemble the total energies. Besides, though only Muliken notation is supported now, we still want to have a index_convention items in the two-electron integral to make things clear.

dgwvfxn:
The serialization and deserialization logic have been tested to work correctly for electronic structure problems.

dgwvfxn:
For energy_offset value, I suggest to assign floating point 0.0 instead of integer 0 as default. (currently coulomb_repulsion has been assigned 0 to its value).

Chemistry/src/DataModel/ProblemDescription/ElectronicStructureProblem.cs
```diff
@@ -0,0 +1,115 @@
...
+    public class ElectronicStructureProblem : IHasMetadata
+    {
+        #region Metadata
+        
+        public Dictionary<string, object> Metadata { get; set; } = new Dictionary<string, object>();
+
+        public Quantity<double> EnergyOffset { get; set; }
+        public Quantity<double> CoulombRepulsion { get; set; }
+
+        // TODO: document that this is self-consistent field energy.
```
msoeken, line 85: Just pointing out that there are still `TODO` mentions, possibly intentional.

Chemistry/src/DataModel/ProblemDescription/ElectronicStructureProblem.cs
```diff
@@ -0,0 +1,115 @@
...
+        public Quantity<double>? ScfEnergy { get; set; }
+        public Quantity<double>? ScfEnergyOffset { get; set; }
+
+        // TODO: document that this is full configuration interaction energy.
```
msoeken, line 89: Just pointing out that there are still `TODO` mentions, possibly intentional.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.2.cs
```diff
@@ -31,6 +33,70 @@ namespace Microsoft.Quantum.Chemistry.Broombridge
...
+                this Dictionary<string, FermionWavefunction<SpinOrbital>> states
+        ) =>
+            states
+            .Select(state => new V0_2.State
+            {
+                Label = state.Key,
+                Energy = new SimpleQuantity
+                {
+                    Value = state.Value.Energy,
+                    Units = "hartree" // FIXME
```
msoeken, line 86: Just pointing out that there are still `FIXME` mentions, possibly intentional.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.2.cs
```diff
@@ -31,6 +33,70 @@ namespace Microsoft.Quantum.Chemistry.Broombridge
...
+                },
+                Method = state.Value.Method switch
+                {
+                    StateType.SingleConfigurational => V0_2.UpdaterStrings.SingleConfigurational,
+                    StateType.SparseMultiConfigurational => V0_2.UpdaterStrings.SparseMultiConfigurational,
+                    StateType.UnitaryCoupledCluster => V0_2.UpdaterStrings.UnitaryCoupledCluster,
+                    _ => "default"
+                }
+                // FIXME: add Superposition and ClusterOperator.
```
msoeken, line 95: Just pointing out that there are still `FIXME` mentions, possibly intentional.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeSerializer.cs
```diff
@@ -0,0 +1,270 @@
...
+                    }
+                );
+        }
+
+        public static void Serialize(TextWriter writer, IEnumerable<ElectronicStructureProblem> problems)
+        {
+            Serializers.SerializeBroombridgev0_2(
+                new Broombridge.V0_2.Data
+                {
+                    // TODO: fix additional properties by converting IEnumerable<ESP> to
```
msoeken, line 57: Just pointing out that there are still `TODO` mentions, possibly intentional.

Chemistry/src/DataModel/Serialization/LegacyFormats/FciDump.cs
```diff
@@ -0,0 +1,156 @@
...
+            (string, double) FormatTerm(KeyValuePair<OrbitalIntegral, DoubleCoeff> term) =>
+                (
+                    String.Join(" ",
+                        ConvertIndices(
+                            term.Key.OrbitalIndices,
+                            OrbitalIntegral.Convention.Dirac,
+                            OrbitalIntegral.Convention.Mulliken
+                        )
+                        .ToOneBasedIndices()
+                        .Select(i => i.ToString())
```
msoeken, line 126: I think that this line may be deleted.

Chemistry/src/Tools/README.md
```diff
@@ -0,0 +1,54 @@
...
+The Quantum Development Kit chemistry command-line tools are distributed as a .NET Core SDK Global Tool, and thus can be installed using the `dotnet` command:
+
+```
+$ dotnet tool install --global Microsoft.Quantum.Chemistry.Tools
+```
+
+> ⚠ NOTE: If you want to use a pre-release version of the Microsoft.Quantum.Chemistry.Tools package, you'll need to [configure the Quantum Development Kit prerelease feed](https://github.com/microsoft/QuantumLibraries#optional-using-prerelease-versions) before proceeding.
+> Once you have the prerelease feed configured, you can install the Microsoft.Quantum.Chemistry.Tools package by providing an explicit version:
+> ```
+> $ dotnet tool install --global Microsoft.Quantum.Chemistry.Tools --version "0.11.2005.1422-beta"
```
msoeken, line 16: This package number is older than the one used in the code.

cgranade:
Thanks for the detailed review, @msoeken! By way of context, a lot of your comments point our preexisting issues, such that I really appreciate the help identifying where  we can take the opportunity to pay down a bit of technical debt. With respect to TODO and FIXME comments, my intent with this PR was to make a substantive step forward with the feature work, but it's not quite ready for master yet (need unit and integration tests, as well as addressing TODOs and FIXMEs); thanks for raising that as well!

Chemistry/src/DataModel/OrbitalIntegral/Indices.cs
```diff
@@ -0,0 +1,66 @@
...
+                2 => indices.Select(o => o).ToArray(),
+                4 => new int[]
+                {
+                    indices.ElementAt(0),
+                    indices.ElementAt(2),
+                    indices.ElementAt(3),
+                    indices.ElementAt(1)
+                },
+                _ => throw new System.ArgumentException(
+                    $"Got indices [{String.Join(", ", indices)}], but Mulliken convention for not 2 or 4 indices is not defined."
```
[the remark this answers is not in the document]
cgranade: It's slightly confusing here, as this concerns a conversion between different indexing conventions. My understanding (@dgwvfxn, please confirm) is that the Mulliken index convention is not defined except for indices of arity two and four.

Chemistry/src/DataModel/ProblemDescription/ElectronicStructureProblem.cs
cgranade: Thanks for pointing this out, added documentation here.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeSerializer.cs
cgranade: This one is definitely intentional for now; it would be a more significant refactoring still to introduce a problem collection class, but we should do that as soon as possible.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.2.cs
cgranade: Yes, the conversion tool does not yet support serializing initial states.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.2.cs
cgranade: Yes, the deserialization logic currently strips away metadata about units, such that we need to revise the logical data model to include that metadata, or to make explicit the assumption that the logical data model is always in terms of Hartree units. In the latter case, deserialization would then need to convert from other units as appropriate.

Chemistry/src/DataModel/OrbitalIntegral/Indices.cs
```diff
@@ -0,0 +1,66 @@
...
+                2 => indices.Select(o => o).ToArray(),
+                4 => new int[]
+                {
+                    indices.ElementAt(0),
+                    indices.ElementAt(3),
+                    indices.ElementAt(1),
+                    indices.ElementAt(2)
+                },
+                _ => throw new System.ArgumentException(
+                    $"Got indices [{String.Join(", ", indices)}], but Dirac convention for not 2 or 4 indices is not defined."
```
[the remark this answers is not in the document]
cgranade: See above; I think it's best to leave this as convention for now.

Chemistry/src/DataModel/OrbitalIntegral/Indices.cs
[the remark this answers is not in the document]
dgwvfxn: @cgranade Your understanding is correct. Same for Dirac convention.

PR #264, opened 2020-06-11
Transformation-based synthesis to apply state-vector permutation

msoeken:
This is an Q# implementation of the unidirectional variant of the transformation based synthesis approach [Miller, Maslov, Dueck, DAC 2003].

The code is moved from the [reversible-logic-synthesis sample](https://github.com/microsoft/Quantum/tree/master/samples/algorithms/reversible-logic-synthesis).

There are tests for the utility functions and the synthesis function.

I tried to follow code conventions and documentation styles as I saw it in other samples.


Standard/src/Synthesis/TransformationBased.qs
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1: Updated guidance has been to drop "All rights reserved" from copyright headers; this isn't critical for existing code, but is suggested for new files.

```suggestion
// Copyright (c) Microsoft Corporation.
```

Standard/src/Synthesis/TransformationBased.qs
```diff
@@ -0,0 +1,173 @@
...
+    /// The second integer is a bit mask for target lines.  Bit indexes which
+    /// are set correspond to target line indexes.
+    ///
+    /// The bit indexes of both integers must be disjoint.
+    internal newtype MCMTMask = (
+        ControlMask : Int, TargetMask : Int
+    );
+
+
+    /// # Summary
```
cgranade, line 31: I agree with keeping this internal for now, but this would be a good candidate for exposing publicly as a future enhancement.

Standard/src/Synthesis/TransformationBased.qs
```diff
@@ -0,0 +1,173 @@
...
+    internal function GateMasksForAssignment (x : Int, y : Int) : MCMTMask[] {
+        let m01 = x &&& ~~~y;
+        let m10 = y &&& ~~~x;
+        return GateMask(y, m01) + GateMask(x, m10);
+    }
+
+
+    /// # Summary
+    /// Update an output pattern according to gate mask.
+    internal function UpdateOutputPattern (pattern : Int, gateMask : MCMTMask) : Int {
```
cgranade, line 77:
```suggestion
    internal function UpdatedOutputPattern (pattern : Int, gateMask : MCMTMask) : Int {
```

Standard/src/Synthesis/TransformationBased.qs
```diff
@@ -0,0 +1,173 @@
...
+    /// Transformation-based synthesis algorithm.
+    ///
+    /// This procedure implements the unidirectional transformation based
+    /// synthesis approach.  Input is a permutation π over 2ⁿ elements {0, ...,
+    /// 2ⁿ-1}, which represents an 𝑛-variable reversible Boolean function.  The
+    /// algorithm performs iteratively the following steps:
+    ///
+    /// 1. Find smallest 𝑥 such that 𝑥 ≠ π(𝑥) = 𝑦
+    /// 2. Find multiple-controlled Toffoli gates, which applied to the outputs
+    ///    make π(𝑥) = 𝑥 and do not change π(𝑥') for all 𝑥' < 𝑥
```
cgranade, line 139: I'd suggest breaking this summary into a summary and a description, with the description having the math in LaTeX notation (Unicode notation works great in samples, since those may not go through docs generation).
```suggestion
    /// # Summary
    /// Applies a given permutation using transformation-based synthesis.
    ///
    /// # Description
    /// This procedure implements the unidirectional transformation based
    /// synthesis approach.  Input is a permutation $\pi$ over $2^n$ elements
    /// $\{0, \dots, 2^n-1\}$, which represents an $n$-variable reversible Boolean function. 
    /// The algorithm performs iteratively the following steps:
    ///
    /// 1. Find smallest $x$ such that $x \ne \pi(x) = y$.
    /// 2. Find multiple-controlled Toffoli operations, which applied to the outputs
    ///    make $\pi(x) = x$ and do not change $\pi(x')$ for all $x' < x$
```

Standard/src/Synthesis/TransformationBased.qs
```diff
@@ -0,0 +1,173 @@
...
+    /// A list of 𝑛 qubits where the Toffoli gates are being applied to.  Note
+    /// that the algorithm does not apply the gates.  But only prepares the
+    /// Toffoli gates.
+    ///
+    /// # Example
+    /// ```Q#
+    /// using (qubits = Qubit[3]) {
+    ///   ApplyPermutationTransformationBased([0, 2, 1, 3], qubits); // synthesize SWAP operation
+    /// }
+    /// ```
```
cgranade, line 154: I really like this example!
```suggestion
    /// # Example
    /// To synthesize a `SWAP` operation:
    /// ```Q#
    /// using (qubits = Qubit[3]) {
    ///   ApplyPermutationTransformationBased([0, 2, 1, 3], qubits);
    /// }
    /// ```
```

Standard/tests/Synthesis.qs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.Quantum.Tests {
+    open Microsoft.Quantum.Arithmetic;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Diagnostics;
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Math;
+    open Microsoft.Quantum.Measurement;
+    open Microsoft.Quantum.Synthesis;
+
+    operation TransformationBasedSynthesisTest () : Unit {
```
cgranade, line 12: I'd suggest using the new `@Test()` attribute for new tests going forward.
```suggestion
    @Test("QuantumSimulator")
    operation CheckTransformationBasedSynthesis() : Unit {
```

Standard/src/Synthesis/TransformationBased.qs
msoeken: Yes, I think so too. I think I am using it also in some other API for the synthesis namespace, so it might already be convenient to expose it publicly soon.

Standard/src/Synthesis/TransformationBased.qs
msoeken: I will change it in a separate commit, together with references to the function and also rename `UpdatePermutation` to `UpdatedPermutation` for consistency.

Standard/tests/Synthesis.qs
msoeken: Good idea. Will change it in a new commit, trying to use `ToffoliSimulator` instead.

Standard/src/Synthesis/TransformationBased.qs
cgranade: Sounds good, shall we open an enhancement issue to track that, then?

msoeken: Yes, we should. I opened one in #266.

Issue #266, opened 2020-06-13
Function to return indexes of one-bits in an integer


msoeken:
**Is your feature request related to a problem? Please describe.**
I have created an internal function `IntegerBits` in the implementation of `Microsoft.Quantum.Synthesis.ApplyPermutationUsingTransformation`, which might be useful for other algorithms as well. These other algorithms may be tracked in this issue.

**Describe the solution you'd like**
A function `function IntegerBits (value : Int, length : Int) : Int[]` that returns an array of length `length` with indexes corresponding to one-bits in `value` (considering `value` is little-endian).

**Describe alternatives you've considered**
Implementing the function internally:
```q#
internal function IntegerBits (value : Int, length : Int) : Int[] {
    return Where(EqualB(true, _), IntAsBoolArray(value, length));
}
```

**Additional context**
These operations make directly or indirectly use of this function:
* Microsoft.Quantum.Synthesis.ApplyPermutationUsingTransformation

PR #267, opened 2020-06-13
Apply single transposition to state-vector

msoeken:
This operation swaps the amplitude at two indexes in the state-vector of a given qubit register.

The operation also makes use of the internal `IntegerBits` function discussed in #266.


Standard/src/Synthesis/Transposition.qs
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Synthesis {
```
cgranade, line 3:
```suggestion

namespace Microsoft.Quantum.Synthesis {
```

Issue #269, opened 2020-06-14
Function to perform operation controlled on truth table


msoeken:
I'd like to discuss API design for a function that controls an operation based on a Boolean function represented by a truth table, similar to the functions `ControlledOnInt` and `ControlledOnBitString`. The truth table can be represented by a `BigInt`, the number of variables can be obtained from the number of controls. (Truth table representations are also explained in the [TruthTables kata](https://github.com/microsoft/QuantumKatas/tree/master/TruthTables). Examples are

```q#
(ControlledOnTruthTable(0xEL, X))([a, b], c);
(ControlledOnTruthTable(0xE8L, T))([a, b, c], d);
```

The first example applies `X` if and only if `a OR b` is true, the second example applies `T`, if and only if the majority of `a`, `b`, and `c` is true.

An implementation in which the truth table is represented as an array of Booleans and in which the operation is `X` can be found [in this oracle-synthesis sample](https://github.com/microsoft/Quantum/blob/master/samples/algorithms/oracle-synthesis/OracleSynthesis.qs#L164). Let's call this operation `ControlledXOnTruthTable`, assuming that the truth table is represented as `BigInt`. Then one can use this implementation to implement the generic operation by conjugating it with `ControlledXOnTruthTable` computing the result on a helper qubit and then controlling the generic operation on that helper qubit. 

However, this implementation is not favourable for the case in which the operation is `X` since the number of operations is doubled (a bit can saved when using measurement-based uncomputation). We cannot make a condition on the operation to choose a special implementation for the case in which the operation is `X`.

The following solutions are possible:

1. Implementing `ControlledOnTruthTable` and `ControlledXOnTruthTable`. `ControlledOnTruthTable` using `X` as operator would have the same behaviour as `ControlledXOnTruthTable` but a most costly implementation.

2. Implementing only `ControlledXOnTruthTable` and give an example in the documentation of this operation how to implement a controlled operation on an arbitrary operation.

3. Implementing `ControlledOnTruthTable` as public API but `ControlledXOnTruthTable` as hidden API, then using AST transformation to replace occurrences of `ControlledOnTruthTable` with operation `X` by the more efficient `ControlledXOnTruthTable` operation.

guenp:
[...]

PR #271, opened 2020-06-16
Apply X controlled on truth table

msoeken:
This PR integrates code from the [oracle synthesis sample](https://github.com/microsoft/Quantum/tree/master/samples/algorithms/oracle-synthesis) into the `Microsoft.Quantum.Synthesis` namespace. The implementation is based on *Construction 1* in [this paper](https://arxiv.org/pdf/2005.12310.pdf).

TODO list:
* [x] Implement variant for arbitrary target
* [x] Implement variant for clean target
* [x] Use `R1Frac` gates instead of `RFrac` gates
* [x] Implement controlled variant
* [x] Docs

API design is discussed in #269

### DIFF
--- a/Standard/src/Synthesis/ControlledOnTruthTable.qs
+++ b/Standard/src/Synthesis/ControlledOnTruthTable.qs
@@ -1,0 +1,137 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Synthesis {
+    open Microsoft.Quantum.Arrays;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Convert;
+    open Microsoft.Quantum.Diagnostics;
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Math;
+
+    /// # Summary
+    /// Computes Hadamard transform of a Boolean function in {-1,1} encoding
+    /// using Yates's method
+    ///
+    /// # Input
+    /// ## func
+    /// Truth table in {-1,1} encoding
+    ///
+    /// # Output
+    /// Spectral coefficients of the function
+    ///
+    /// # Example
+    /// ```Q#
+    /// FastHadamardTransform([1, 1, 1, -1]); // [2, 2, 2, -2]
+    /// ```
+    ///
+    /// # Reference
+    /// Frank Yates: The design and analysis of factorial experiments, in:
+    /// Technical Communication No. 35, Imperial Bureau of Soil Science,
+    /// London (1937)
+    internal function FastHadamardTransform(func : Int[]) : Int[] {
+        let bits = BitSizeI(Length(func) - 1);
+        mutable res = func;
+        for (m in 0..bits - 1) {
+            mutable s = 1 <<< m;
+            for (i in 0..(2 * s)..Length(func) - 1) {
+                mutable k = i + s;
+                for (j in i..i + s - 1) {
+                    mutable t = res[j];
+                    set res w/= j <- res[j] + res[k];
+                    set res w/= k <- t - res[k];
+                    set k = k + 1;
+                }
+            }
+        }
+        return res;
+    }
+
+    /// # Summary
+    /// Extends a spectrum by inverted coefficients
+    ///
+    /// # Input
+    /// ## spectrum
+    /// Spectral coefficients
+    ///
+    /// # Output
+    /// Coefficients followed by inverted copy
+    ///
+    /// # Example
+    /// ```Q#
+    /// Extend([2, 2, 2, -2]); // [2, 2, 2, -2, -2, -2, -2, 2]
+    /// ```
+    internal function Extend(spectrum : Int[]) : Int[] {
+        return spectrum + Mapped(NegationI, spectrum);
+    }
+
+    /// # Summary
+    /// {-1,1} coding of a Boolean truth value
+    ///
+    /// # Input
+    /// ## b
+    /// Boolean value
+    ///
+    /// # Output
+    /// 1, if `b` is false, otherwise -1
+    internal function RMEncoding(b : Bool) : Int {
+        return b ? -1 | 1;
+    }
+
+    /// # Summary
+    /// Encode truth table in {1,-1} coding
+    ///
+    /// # Input
+    /// ## table
+    /// Truth table as array of truth values
+    ///
+    /// # Output
+    /// Truth table as array of {1,-1} integers
+    ///
+    /// # Example
+    /// ```Q#
+    /// Encode([false, false, false, true]); // [1, 1, 1, -1]
+    /// ```
+    internal function Encode(table : Bool[]) : Int[] {
+        return Mapped(RMEncoding, table);
+    }
+
+    internal function SizeAdjustedTruthTable(table : Bool[], numVars : Int) : Bool[] {
+        let numEntries = 2^numVars;
+        if (numEntries < Length(table)) {
+            return table[...numEntries - 1];
+        } elif (numEntries > Length(table)) {
+            return Padded(numEntries, false, table);
+        } else {
+            return table;
+        }
+    }
+
+    operation ControlledXOnTruthTable (func : BigInt, controlRegister : Qubit[], targetRegister : Qubit) : Unit {
+        let vars = Length(controlRegister);
+
+        let maxValue = PowL(2L, 2^vars);
+        Fact(func >= 0L and func < maxValue, $"Argument func must be value from 0 to {maxValue}");
+
+        let tt = BigIntAsBoolArray(func);
+        let table = Encode(SizeAdjustedTruthTable(BigIntAsBoolArray(func), vars));
+        let spectrum = Extend(FastHadamardTransform(table));
+
+        let qubits = controlRegister + [targetRegister];
+
+        HY(targetRegister);
+
+        for (i in 0..vars) {
+            let start = 1 <<< i;
+            let code = GrayCode(i);
+            for (j in 0..Length(code) - 1) {
+                let (offset, ctrl) = code[j];
+                RFrac(PauliZ, -spectrum[start + offset], vars + 2, qubits[i]);
+                if (i != 0) {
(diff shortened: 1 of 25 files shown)


Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,193 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+namespace Microsoft.Quantum.Synthesis {
```
cgranade, line 3:
```suggestion

namespace Microsoft.Quantum.Synthesis {
```

cgranade, line 3: At some point, we should add a `Properties/NamespaceInfo.qs` to the `Microsoft.Quantum.Synthesis` namespace.

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,193 @@
...
+    ///
+    /// # Output
+    /// Truth table as array of {1,-1} integers
+    ///
+    /// # Example
+    /// ```Q#
+    /// Encode([false, false, false, true]); // [1, 1, 1, -1]
+    /// ```
+    internal function Encode(table : Bool[]) : Int[] {
+        return Mapped(RMEncoding, table);
```
cgranade, line 95: This is a great example of where lambda functions and operations (https://github.com/microsoft/qsharp-compiler/issues/181) could really help.

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,193 @@
...
+                    if (i != 0) {
+                        CNOT(qubits[ctrl], qubits[i]);
+                    }
+                }
+            }
+
+            H(targetRegister);
+        }
+        adjoint self;
+    }
```
cgranade: Since `adjoint self` is the default for `is Adj`, it may be more clear to leave that implicit by omitting the `body` declaration.
```suggestion
    operation ControlledXOnTruthTable (func : BigInt, controlRegister : Qubit[], targetRegister : Qubit)
    : Unit is Adj {
        let vars = Length(controlRegister);

        let maxValue = PowL(2L, 2^vars);
        Fact(func >= 0L and func < maxValue, $"Argument func must be value from 0 to {maxValue}");

        let tt = BigIntAsBoolArray(func);
        let table = Encode(SizeAdjustedTruthTable(BigIntAsBoolArray(func), vars));
        let spectrum = Extend(FastHadamardTransform(table));

        let qubits = controlRegister + [targetRegister];

        HY(targetRegister);

        for (i in 0..vars) {
            let start = 1 <<< i;
            let code = GrayCode(i);
            for (j in 0..Length(code) - 1) {
                let (offset, ctrl) = code[j];
                R1Frac(spectrum[start + offset], vars + 1, qubits[i]);
                if (i != 0) {
                    CNOT(qubits[ctrl], qubits[i]);
                }
            }
        }

        H(targetRegister);
    }
```

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,193 @@
...
+
+    operation ControlledXOnTruthTableWithCleanTarget (func : BigInt, controlRegister : Qubit[], targetRegister : Qubit) : Unit {
+        body (...) {
+            let vars = Length(controlRegister);
+
+            let maxValue = PowL(2L, 2^vars);
```
cgranade, line 155: Would a bit-shift here be helpful?
```suggestion
            let maxValue = 1 <<< 2^vars;
```

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,193 @@
...
+
+            H(targetRegister);
+            AssertProb([PauliZ], [targetRegister], One, 0.5, "Probability of the measurement must be 0.5", 1e-10);
+
+            if (IsResultOne(M(targetRegister))) {
+                for (i in 0..vars - 1) {
+                    let start = 1 <<< i;
+                    let code = GrayCode(i);
+                    for (j in 0..Length(code) - 1) {
+                        let (offset, ctrl) = code[j];
```
cgranade, line 192:
```suggestion
                    for ((j, (offset, ctrl)) in Enumerated(code)) {
```

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: I wasn't aware of this. I thought that `Adj` will automatically reverse and mirrors the behaviour of `adjoint inverse`, but `adjoint self` explicitly states that the order of execution does not make a difference (see, e.g., in the case of [CX](https://github.com/microsoft/QuantumLibraries/blob/master/Standard/src/Canon/CommonGates.qs#L42))

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: I didn't realise at that point that I can do a left-shift with BigInt on LHS and Int on RHS. So I'll change it to `let maxValue = 1L <<< 2^vars;`

msoeken:
Thanks for the comments, @cgranade. Will address them in a separate commit, because my local version converged already a bit.

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: I really like this one!

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: Will address this in the next PR to have a clear separation.

Standard/src/Synthesis/ControlledOnTruthTable.qs
cgranade: Sorry, I had misread your code there; you're right that `inverse` is the default. Please disregard.

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,271 @@
...
+    /// ## target
+    /// Target qubit
+    ///
+    /// # See Also
+    /// - @"microsoft.quantum.synthesis.controlledxontruthtablewithcleantarget"
+    ///
+    /// # References
+    /// - [*N. Schuch*, *J. Siewert*, PRL 91, no. 027902, 2003, arXiv:quant-ph/0303063](https://arxiv.org/abs/quant-ph/0303063)
+    /// - [*Mathias Soeken*, *Martin Roetteler*, arXiv:2005.12310](https://arxiv.org/abs/2005.12310)
+    operation ControlledXOnTruthTable (func : BigInt, controlRegister : Qubit[], target : Qubit) : Unit {
```
cgranade, line 158: Should we call this `ApplyXControlledOnTruthTable`, perhaps?

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /// # Summary
+    /// Applies the `X` operation on `target`, if the Boolean function `func` evaluates
+    /// to true for the classical assignment in `controlRegister`.  In other words, it implements
+    /// the unitary operation
+    /// \begin{align}
+    ///    U\ket{x}\ket{y} = \ket{x}\ket{y \oplus f(x)}
+    /// \end{align}
+    /// where $x$ and $y$ represent `controlRegister` and `target`, respectively.
```
cgranade, line 132: I'd suggest splitting this into a summary and a description, so that the summary can be stated without mathematical notation. Not all clients support LaTeX in summaries, but that's extremely useful content to have in the description.

It may also be worth making `X` a link, e.g.: `@"microsoft.quantum.intrinsic.x"`.

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,271 @@
...
+    /// # Input
+    /// ## func
+    /// Boolean truth table represented as big integer
+    /// ## controlRegister
+    /// Register of control qubits
+    /// ## target
+    /// Target qubit (must be in $\ket{0}$ state)
+    ///
+    /// # See Also
+    /// - @"microsoft.quantum.synthesis.controlledxontruthtable"
```
cgranade, line 216:
```suggestion
    /// - Microsoft.Quantum.Synthesis.ControlledXOnTruthTable
```

Standard/src/Synthesis/ControlledOnTruthTable.qs
```diff
@@ -0,0 +1,271 @@
...
+    operation ControlledXOnTruthTableWithCleanTarget (func : BigInt, controlRegister : Qubit[], target : Qubit) : Unit {
+        body (...) {
+            let vars = Length(controlRegister);
+
+            let maxValue = PowL(2L, 2^vars);
+            Fact(func >= 0L and func < maxValue, $"Argument func must be value from 0 to {maxValue}");
+            AssertAllZero([target]);
```
cgranade, line 223:
```suggestion
            AssertQubit(Zero, target);
```

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: Yes, sounds good. I wanted to make them sound similar to `ControlledOnBitString` etc., but due to the limitation to `X` and the signature they are in fact different. Maybe we can later add a more generic operation `ControlledOnTruthTable` that can make use of this operation.

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: Good to know.

Standard/src/Synthesis/ControlledOnTruthTable.qs
msoeken: Resolved in commit

PR #272, opened 2020-06-16
Chemistry conversion unit and integration tests

cgranade:
This PR adds new tests to check that the improved serialization support in the chemistry data model are correct, and to guard against regression.


Chemistry/src/DataModel/Serialization/LegacyFormats/FciDump.cs
```diff
@@ -85,6 +85,11 @@ public static IEnumerable<ElectronicStructureProblem> Deserialize(TextReader rea
                 )
                 .Distinct()
             );
+
+            // The identity term in deserialized Hamiltonians is the sum of the 
+            // Coloumb repulsion and the energy offset. Since only the former
+            // exists in FCIDUMP, we set the identity term accordingly.
+            hamiltonian.Add(new OrbitalIntegral(), coulomb);
```
cgranade, line 92: This line fixes a bug discovered through the new unit tests.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.2.cs
```diff
@@ -95,6 +95,62 @@ internal static class BroombridgeExtensionsV0_2
...
+        {
+            var wavefunctions = new Dictionary<string, Fermion.FermionWavefunction<OrbitalIntegrals.SpinOrbital>>();
+            foreach (var initialState in initialStates ?? new List<V0_2.State>())
+            {
+                var finalState = new FermionWavefunction<SpinOrbital>();
+
+                var (method, energy, outputState) = V0_2.ToWavefunction(initialState);
+
+                var setMethod = V0_2.ParseInitialStateMethod(initialState.Method);
+                var setEnergy = energy;
```
msoeken, line 112: I don't see the need for this temporary variable. Why not referring to `energy` instead. Variable `setEnergy` is not written to after this assignment.

Chemistry/tests/DataModelTests/Extensions.cs
```diff
@@ -0,0 +1,123 @@
...
+                throw new AggregateException(exceptions);
+            }
+        }
+
+        internal static Action HasUnits<T>(this Quantity<T> quantity, string units) =>
+            () => Assert.Equal(quantity.Units, units);
+
+        internal static IEnumerable<Action> IsEqualTo(this ElectronicStructureProblem expected, ElectronicStructureProblem actual)
+        {   
+            // TODO: check rest of metadata.
```
msoeken, line 59: Just pointing out to the `TODO` as this PR does not seem to be work-in-progress.

cgranade: Thanks for catching this; in this case, the test is incomplete because this PR does not yet resolve the missing functionality needed to serialize initial state suggestions to Broombridge. Currently, initial state suggestions can be deserialized, but cannot be written out, such that any electronic structure data containing such suggestions will not round-trip correctly. I'll add a feature request to track that work and link it from this comment so as to be more explicit.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.2.cs
cgranade: Agreed, this was pre-existing code that was moved with a refactor, but I agree with fixing it in this PR to be more readable and maintainable. Thank you!

PR #273, opened 2020-06-16
Fix #262.

### LINKED ISSUE #262: `ControlledOnBitString` behavior when bitstring and control register are not the same length not documented
**Describe the bug**
This _may_ not be a bug, but I wanted to capture that the behavior for `ControlledOnBitString` is not clear when the control register it is applied to is longer than the control bitstring (the first argument) itself.
(ControlledOnBitString(address, ApplyPauliFromBitString(PauliX, true, _, _)))
            (addressRegister!, (values, targetRegister));
**To Reproduce**
See [gist](https://gist.github.com/crazy4pi314/516936ac7777660d2694eb9d7f957263) here, run the sample in binder: [Binder](https://mybinder.org/v2/gist/crazy4pi314/516936ac7777660d2694eb9d7f957263/master)

```qsharp

operation ControllingStringCorrectLength() : Unit {
    let address = [true, false, true];
    using((addressRegister, target) = (Qubit[3], Qubit())){
        // Prepare address register to address value
        ApplyPauliFromBitString(PauliX, true, address, addressRegister);
        // Control on the same address, should put target qubit in superposition.
        (ControlledOnBitString(address, H))
            (addressRegister, target);
        DumpMachine();
        ResetAll(addressRegister+[target]);
    }
}

operation ControllingStringShort() : Unit {
    let address = [true, false];
    using((addressRegister, target) = (Qubit[3], Qubit())){
        // Prepare address register to address value
        ApplyPauliFromBitString(PauliX, true, address, addressRegister);
        // Control on the same address, should put target qubit in superposition.
        (ControlledOnBitString(address, H))
            (addressRegister, target);
        DumpMachine();
        ResetAll(addressRegister+[target]);
    }
}

```

**Expected behavior**
I guess I am not sure what I should expect in this case? That is part of why I am filing this :P 

**Screenshots**
Screenshot of the attached 
image

**System information**
 - OS: Windows 10
image

cgranade:
This PR fixes #262 by leaving "extra" qubits in the control register explicitly uncontrolled.


Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -21,16 +21,16 @@ namespace Microsoft.Quantum.Canon {
     /// # Remarks
     /// The length of `bits` and `controlRegister` must be equal.
```
msoeken, line 22: This comment is misleading in light of the new implementation.

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -21,16 +21,16 @@ namespace Microsoft.Quantum.Canon {
...
-    operation ControlledOnBitStringImpl<'T> (bits : Bool[], oracle : ('T => Unit is Adj + Ctl), controlRegister : Qubit[], targetRegister : 'T) : Unit
-    {
-        body (...)
-        {
-            ApplyWithCA(ApplyPauliFromBitString(PauliX, false, bits, _), Controlled oracle(_, targetRegister), controlRegister);
+    operation ControlledOnBitStringImpl<'T> (bits : Bool[], oracle : ('T => Unit is Adj + Ctl), controlRegister : Qubit[], targetRegister : 'T)
+    : Unit is Adj + Ctl {
+        // Use a subregister of the controlled register when
+        // bits is shorter than controlRegister.
+        let controlSubregister = controlRegister[...Length(bits) - 1];
```
msoeken, line 28: What happens if `Length(bits) > Length(controlRegister)`?

Standard/src/Canon/Utils/ControlledOnBitString.qs
```diff
@@ -60,7 +60,9 @@ namespace Microsoft.Quantum.Canon {
     /// A unitary operation that applies `oracle` on the target register if the control register state corresponds to the bit mask `bits`.
     ///
     /// # Remarks
-    /// The length of `bits` and `controlRegister` must be equal.
+    /// The pattern given by `bits` may be shorter than `controlRegister`,
```
msoeken, line 63: What happens if `bits` is longer than `controlRegister`? If it's not allowed we should raise an error in that case.

Standard/src/Canon/Utils/ControlledOnBitString.qs
cgranade: Indeed, I thought I had caught that one. Thanks for pointing it out.

Standard/src/Canon/Utils/ControlledOnBitString.qs
cgranade: Sounds good, will make sure to add that explicitly.

msoeken:
> > Thanks @cgranade, all comments are addressed. I was about to use the (now internal) operation `ControlledOnBitStringImpl` to draw inverted controls in the next blog post. I am wondering whether this is still possible if the Q# operation is marked `internal`.
> 
> The previous implementation predated `internal`, but the docs suppress coverage of anything ending with `Impl` as that was one of our legacy methods for denoting an internal-only callable. Thus, the `ControlledOnBitStringImpl` operation was already internal-only. That said, we could make it public now that it has a name that keeps with our API design principles; that would just take adding more API documentation comments, but that's about it.

I support the idea of having `ApplyControlledOnBitString` and `ApplyControlledOnInt` public API. For some users it may be more intuitive than `ControlledOnBitString` and `ControlledOnInt`. There'll be no code duplication. We can refer from the `Apply...` operations to the `ControlledOn...` functions via *See also* links.

cgranade:
> I support the idea of having `ApplyControlledOnBitString` and `ApplyControlledOnInt` public API. For some users it may be more intuitive than `ControlledOnBitString` and `ControlledOnInt`. There'll be no code duplication. We can refer from the `Apply...` operations to the `ControlledOn...` functions via _See also_ links.

Sounds good, let me get on that, then.

PR #276, opened 2020-06-17
Add Repeat and DrawMany operations.

### LINKED ISSUE #245: Operations to repeat a call 𝑛 times
**Is your feature request related to a problem? Please describe.**
Following offline discussions with @alan-geller and @tcNickolas, and following a suggestion from @crazy4pi314, it would be nice to have operations which repeatedly apply another operation 𝑛 times.

**Describe the solution you'd like**
Following the example of R, this feature request would introduce five new operations into the Microsoft.Quantum.Standard package:

```Q#
namespace Microsoft.Quantum.Canon {
    operation Repeat<'TInput>(op : ('TInput => Unit), nTimes : Int, input : 'TInput) : Unit { ... }
    operation RepeatA<'TInput>(op : ('TInput => Unit is Adj), nTimes : Int, input : 'TInput) : Unit is Adj { ... }
    operation RepeatC<'TInput>(op : ('TInput => Unit is Ctl), nTimes : Int, input : 'TInput) : Unit is Ctl { ... }
    operation RepeatCA<'TInput>(op : ('TInput => Unit is Adj + Ctl), nTimes : Int, input : 'TInput) : Unit is Adj + Ctl { ... }
}

namespace Microsoft.Quantum.Arrays {
    operation Replicate<'TInput, 'TOutput>(op : ('TInput => 'TOutput), nTimes : Int, input : 'TInput) : 'TOutput[] { ... }
}
```

**Describe alternatives you've considered**
The following currently work in place of the Repeat and Replicate operations proposed above, but are unwieldy:

```Q#
ApplyToEach(Delay(op, input, _), ConstantArray(nTimes, ()));
ForEach(Delay(op, input, _), ConstantArray(nTimes, ()));
```

Alternative names considered for `Replicate`:

- `Repeat`: doesn't make clear that outputs are collected.
- `Sample`: may work, but is too close to Microsoft.Quantum.MachineLearning.Sampled; the noun-vs-verb distinction is not held to uniformly enough to make the distinction clear.
- `DrawSample`: may work, but verb "draw" is a bit esoteric.
- `DoNTimes`: similar to `Repeat`, doesn't make clear that outputs are collected.
- `Collect`: doesn't make clear what is being collected.
- `RepeatAndCollect`: somewhat clear, but unnecessarily verbose

cgranade:
This PR fixes #245 by adding new `Repeat` and `DrawMany` operations to the Q# standard libraries.

This is still in draft pending the following TODOs:
- [ ] Add tests.

### DIFF
--- a/Standard/src/Canon/Repeat.qs
+++ b/Standard/src/Canon/Repeat.qs
@@ -21,8 +21,8 @@ namespace Microsoft.Quantum.Canon {
     /// # Example
     /// The following are equivalent:
     /// ```Q#
-    /// Repeat(H, 2, target);
-    /// NoOp<Qubit>(target);
+    /// Repeat(U, 17, target);
+    /// (Bound(ConstantArray(17, U)))(target);
     /// ```
     ///
     /// # See Also
@@ -54,8 +54,8 @@ namespace Microsoft.Quantum.Canon {
     /// # Example
     /// The following are equivalent:
     /// ```Q#
-    /// Repeat(H, 2, target);
-    /// NoOp<Qubit>(target);
+    /// RepeatA(U, 17, target);
+    /// (BoundA(ConstantArray(17, U)))(target);
     /// ```
     ///
     /// # See Also
@@ -87,8 +87,8 @@ namespace Microsoft.Quantum.Canon {
     /// # Example
     /// The following are equivalent:
     /// ```Q#
-    /// Repeat(H, 2, target);
-    /// NoOp<Qubit>(target);
+    /// RepeatC(U, 17, target);
+    /// (BoundC(ConstantArray(17, U)))(target);
     /// ```
     ///
     /// # See Also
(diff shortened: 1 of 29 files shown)


Standard/src/Canon/Repeat.qs
```diff
@@ -0,0 +1,138 @@
...
+    ///
+    /// # Type Parameters
+    /// ## TInput
+    /// The type of input expected by `op`.
+    ///
+    /// # Example
+    /// The following are equivalent:
+    /// ```Q#
+    /// Repeat(H, 2, target);
+    /// NoOp<Qubit>(target);
```
msoeken, line 25: I suggest to use a different example, or introduce it differently. These are not really equivalent, e.g., when performing resource estimation or visualisation. The example therefore may suggest that `Repeat` is performing cancellation.

cgranade: Good point, thanks. Will address that and write a better example.

PR #285, opened 2020-06-21
Creating Chemistry Metapackage

anpaz-msft:
Currently, the Chemistry package adds reference to two dlls form two different projects:
- DataModel: with the C# classes required to parse broombridge files and create the necessary Q# structures.
- Runtime: With the Q# code to run the actual simulations
Having a single nuget package makes it hard to break dependencies and to reprlace package with project references for the right environment.
This change creates a nuget package for each project, and converts the Microsoft.Quantum.Chemistry package to a metapackage that simply refers the other two.


Chemistry/src/Metapackage/Metapackage.csproj
```diff
@@ -0,0 +1,36 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
+    <PlatformTarget>x64</PlatformTarget>
+    <AssemblyName>Microsoft.Quantum.Chemistry.Metapackage</AssemblyName>
```
cgranade, line 6: The trick I used in Quantum-NC was to use the `<IncludeBuildOutput>false</IncludeBuildOutput>` property in the metapackage, such that no blank assembly is needed. Would the same trick work here instead of setting the assembly name and signing an empty assembly?

cgranade:
Ugh, I hate conflicts in `.sln` files. It looks like the main source of conflict is adding the conversion tool, so let me revise that manually and update this PR. Thanks for your patience.

cgranade:
My apologies, @anpaz-msft, I accidentally pushed a commit past the merge conflict resolution.

PR #292, opened 2020-07-06
Simulate ApplyAnd and ApplyLowDepthAnd classically

msoeken:
This PR overrides the behaviour of `ApplyAnd` and `ApplyLowDepthAnd` to use the `CCNOT` behaviour when used with `ToffoliSimulator`.

### DIFF
--- a/Standard/src/Canon/And.cs
+++ b/Standard/src/Canon/And.cs
@@ -10,11 +10,11 @@ using Microsoft.Quantum.Simulation.Simulators;
 
 namespace Microsoft.Quantum.Canon
 {
-    /// <summary>
-    /// Uses CCNOT implementation when executed using ToffoliSimulator.
-    /// </summary>
     public partial class ApplyAnd
     {
+        /// <summary>
+        /// Uses CCNOT implementation when executed using ToffoliSimulator.
+        /// </summary>
         public class Native : ApplyAnd
         {
             private ToffoliSimulator? simulator = null;
@@ -45,15 +45,15 @@ namespace Microsoft.Quantum.Canon
         }
     }
 
-    /// <summary>
-    /// Uses CCNOT implementation when executed using ToffoliSimulator.
-    /// </summary>
     public partial class ApplyLowDepthAnd
     {
+        /// <summary>
+        /// Uses CCNOT implementation when executed using ToffoliSimulator.
+        /// </summary>
         public class Native : ApplyLowDepthAnd
         {
-            private ToffoliSimulator simulator = null;
-            protected CCNOT CCNOT { get; set; } = null;
+            private ToffoliSimulator? simulator = null;
+            protected CCNOT? CCNOT { get; set; } = null;
 
             public Native(IOperationFactory m) : base(m)
             {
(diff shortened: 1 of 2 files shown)


Standard/src/Canon/And.cs
```diff
@@ -0,0 +1,80 @@
...
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.Quantum.Intrinsic;
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators;
+
+namespace Microsoft.Quantum.Canon
+{
+    /// <summary>
```
cgranade, line 11: Technically, this comment as currently written applies to the whole `ApplyAnd` class, and not to the native implementation at Line 16. I'd suggest moving this comment down to make it more clear what the comment addresses.

Standard/src/Canon/And.cs
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
```
cgranade, line 3:
```suggestion

#nullable enable

```

Standard/src/Canon/And.cs
```diff
@@ -0,0 +1,80 @@
...
+{
+    /// <summary>
+    /// Uses CCNOT implementation when executed using ToffoliSimulator.
+    /// </summary>
+    public partial class ApplyAnd
+    {
+        public class Native : ApplyAnd
+        {
+            private ToffoliSimulator simulator = null;
+            protected CCNOT CCNOT { get; set; } = null;
```
cgranade, line 19:
```suggestion
            private ToffoliSimulator? simulator = null;
            protected CCNOT? CCNOT { get; set; } = null;
```

Standard/src/Canon/And.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            public override Func<(Qubit, Qubit, Qubit), QVoid> AdjointBody => CCNOT?.AdjointBody ?? base.AdjointBody;
+
+            public override Func<(IQArray<Qubit>, (Qubit, Qubit, Qubit)), QVoid> ControlledBody => CCNOT?.ControlledBody ?? base.ControlledBody;
+
+            public override Func<(IQArray<Qubit>, (Qubit, Qubit, Qubit)), QVoid> ControlledAdjointBody => CCNOT?.ControlledAdjointBody ?? base.ControlledAdjointBody;
+        }
+    }
+
+    /// <summary>
```
cgranade, line 46: The same comments as above apply here as well.

Standard/src/Canon/And.cs
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

Standard/src/Canon/And.cs
msoeken: Thanks. I tried this in the first place, but had some problems with the `as` cast. Now I see that I forgot to make `ToffoliSimulator` explicitly nullable.

Standard/src/Canon/And.cs
msoeken: Will move it lower.

msoeken:
@cgranade, thanks for your quick feedback and helpful comments.

cgranade:
No worries, happy to help!

PR #297, opened 2020-07-18
Replacing icon for packages that were missed in the recent asset update.

ricardo-espinoza:
As a continuation of #295 , replacing the icon for three packages that were missed in the original asset update.


ricardo-espinoza:
As a continuation of #295 , replacing the icon for three packages that were missed in the original asset update.

PR #302, opened 2020-07-30
Improvements to M.Q.Diagnostics namespace

cgranade:
This PR (still in draft for early feedback and API review) provides a number of improvements to the Microsoft.Quantum.Diagnostics namespace in the Microsoft.Quantum.Standard package:

- **New visualization support**
  This PR provides a new package, Microsoft.Quantum.Standard.Visualization, that uses the runtime emulation feature to interpret diagnostics sent to the `MaybeDisplayDiagnostic` event, allowing for improved diagnostics support in IQ#. This architecture reduces cross-repo dependencies, as visualizations can be provided in library runtime emulation classes, without requiring modifications to the runtime or to IQ#.
- **New `DumpOperation` callable**
  This PR provides a new operation, `DumpOperation`, that uses the Choi–Jamiłkowski isomorphism and `DumpRegister` together to extract a unitary representation of adjointable operations and to display this representation as a diagnostic. On machines that do not support this representation, but that support the `H` and `CNOT` intrinsics and `DumpMachine`, the Choi–Jamiłkowski state for the given operation is shown instead. On machines that support neither, the operation can be safely ignored.
  
image

- **New `Allow*` callables**
  This PR also provides operations using a new verb, `Allow*`, representing that a condition must hold inside of the `apply` arm of a `within`/`apply` block. For example, this functionality can in principle be used in some oracle-based katas by enforcing that a given oracle operation is not called more than once within a given `apply`. Similarly, `AllowAtMostNQubits` fails when more than a given number of qubits is allocated inside of an `apply` block.

_Remaining work:_
- [ ] API review
- [x] C# documentation comments
- [x] Unit testing
    NB: unit testing may require the legacy Q# unit testing framework, so as to allow for `ShouldFail` tests.
- [x] Remove extension method placeholder for microsoft/iqsharp#239 by updating to latest beta.

+@tcNickolas for possible application to oracle-based katas, +@rmshaffer for IQ# impact.
+@guenp and @msoeken for API review.


Standard/src/Diagnostics/Emulation/Internal.cs
```diff
@@ -0,0 +1,93 @@
...
+            Data[(int)idx, 1] = img;
+            return true;
+        }
+
+        public override bool Dump(IQArray<Qubit>? qubits = null)
+        {
+            var count = qubits == null
+                        ? this.Simulator.QubitManager!.GetAllocatedQubitsCount()
+                        : qubits.Length;
+            var nQubitsPerRegister = ((int) count / 2);
```
msoeken, line 40:
```suggestion
            var nQubitsPerRegister = ((int)count / 2);
```

Seems that `count` is cast, but it looks like `count / 2` is cast because of the space.

Standard/src/Diagnostics/Emulation/AllowOperationCalls.cs
```diff
@@ -0,0 +1,106 @@
...
+            public Native(IOperationFactory m) : base(m)
+            {
+                Simulator = m as SimulatorBase;
+            }
+
+            public override Func<(long, IUnitary), QVoid> Body => (_args) =>
+            {
+                if (Simulator == null) return QVoid.Instance;
+
+                (var nTimes, var op) = _args;
```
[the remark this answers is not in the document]
cgranade: I definitely prefer it as `var (nTimes, op)`, agreed. At one point, intelliSense yelled at me for doing it that way, so I switched to `(var nTimes, var op)`, but that may have been a misconfiguration on my part.

cgranade:
Looks like the test process crashed, will rerun on the next commit.

Standard/src/Diagnostics/Emulation/Extensions.cs
```diff
@@ -0,0 +1,77 @@
...
+        internal static SimulatorEventDisposer RegisterOperationHandlers(
+                this SimulatorBase simulator,
+                Action<ICallable, IApplyData> startOperation,
+                Action<ICallable, IApplyData>? endOperation = null
+        ) => new SimulatorEventDisposer(
+            simulator,
+            startOperation,
+            endOperation == null
+                ? (callable, data) => {}
+                : endOperation
```
msoeken, line 74:
```suggestion
            endOperation ?? (callable, data) => {}
```

Standard/src/Diagnostics/Emulation/Internal.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace Microsoft.Quantum.Diagnostics
+{
+    internal class ArrayDumper : QuantumSimulator.StateDumper
+    {
+        // NB: NumSharp does not yet support complex numbers, so we store data
+        //     as an array with a trailing index of length 2.
+        internal NumSharp.NDArray? Data = null;
+        public ArrayDumper(QuantumSimulator sim) : base(sim)
+        {
+
```
msoeken, line 24:
```suggestion
```

Standard/src/Diagnostics/Emulation/Internal.cs
```diff
@@ -0,0 +1,93 @@
...
+            Data[(int)idx, 0] = real;
+            Data[(int)idx, 1] = img;
+            return true;
+        }
+
+        public override bool Dump(IQArray<Qubit>? qubits = null)
+        {
+            var count = qubits == null
+                        ? this.Simulator.QubitManager!.GetAllocatedQubitsCount()
+                        : qubits.Length;
```
msoeken, line 39:
```suggestion
            var count = qubits?.Length ?? Simulator.QubitManager!.GetAllocatedQubitsCount();
```

Standard/src/Diagnostics/Emulation/Internal.cs
```diff
@@ -0,0 +1,93 @@
...
+                    new DisplayableUnitaryOperator
+                    {
+                        Data = arrayDumper.Data,
+                        Qubits = target.ToList()
+                    }
+                );
+                return QVoid.Instance;
+            }
+
+            public override Func<(IQArray<Qubit>, IQArray<Qubit>), QVoid> Body => (__in__) =>
```
msoeken, line 79:
```suggestion
            public override Func<(IQArray<Qubit>, IQArray<Qubit>), QVoid> Body => __in__ =>
```

Standard/tests/Diagnostics/AllowTests.qs
```diff
@@ -0,0 +1,57 @@
...
+                // Should run four times, one more
+                // than the three allowed.
+                for (idx in 0..3) {
+                    X(q);
+                }
+            }
+        }
+    }
+
+    @Diag.Test("QuantumSimulator")
```
msoeken, line 22: Is `@Test` defined in Diagnostics? I thought it was in Core. Or is it moved with this PR?

Visualization/src/CallSiteEncoders.cs
```diff
@@ -0,0 +1,91 @@
...
+        public string MimeType => MimeTypes.PlainText;
+
+        public CallSitesToTextEncoder(ILogger<CallSitesToTextEncoder> logger)
+        {
+            this.logger = logger;
+        }
+
+        public EncodedData? Encode(object displayable)
+        {
+            if (displayable is CallSites sites)
```
msoeken, line 29: This might be a bit more concise with `switch` expression and single-expression return body.

Visualization/src/DisplayableUnitaryEncoders.cs
```diff
@@ -0,0 +1,133 @@
...
+        public string MimeType => MimeTypes.PlainText;
+
+        public DisplayableUnitaryOperatorToTextEncoder(ILogger<DisplayableUnitaryOperatorToTextEncoder> logger)
+        {
+            this.logger = logger;
+        }
+
+        public EncodedData? Encode(object displayable)
+        {
+            if (displayable is DisplayableUnitaryOperator op)
```
msoeken, line 29: Similar comment as in the other `Encode` method.

Visualization/src/CallSiteEncoders.cs
cgranade: Agreed, though there's a slight bug in C# 8 that makes that uglier than it should be in this particular case due to not inferring that `null` means a value of type `Nullable<EncodedData>` and not a null reference. I'll work with this snippet to make it look a bit nicer, though. Thanks!

Standard/tests/Diagnostics/AllowTests.qs
cgranade: Yes, the `Test` attribute is in Microsoft.Quantum.Diagnostics: https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.test

Standard/src/Diagnostics/Emulation/AllowOperationCalls.cs
```diff
@@ -0,0 +1,106 @@
...
+
+                if (!Handlers.ContainsKey(Key))
+                {
+                    Handlers[Key] = new Stack<IDisposable>();
+                }
+
+                bool IsSelf(ICallable callable) =>
+                    callable.FullName == "Microsoft.Quantum.Diagnostics.AllowAtMostNCallsCA";
+
+                var failed = false;
```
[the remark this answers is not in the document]
cgranade: Good catch, added. Thanks!

cgranade:
@anpaz-msft: It looks like the Linux test host is crashing pretty frequently again, would you have any suggestions? Thanks!

```
The active test run was aborted. Reason: Test host process crashed
```

Visualization/src/CallSiteEncoders.cs
msoeken: That's good to know. Thanks for the additional information.

Visualization/src/Extensions.cs
```diff
@@ -0,0 +1,42 @@
...
+            var prefix = Enumerable.Repeat(All, axis).ToArray();
+            foreach (var idx in Enumerable.Range(0, array.Shape[axis]))
+            {
+                Slice[] slice = prefix.Concat(new Slice[] { idx, Ellipsis }).ToArray();
+                yield return array[slice];
+            }
+        }
+
+        // TODO: remove once https://github.com/microsoft/iqsharp/pull/239 is
+        //       resolved.
```
rmshaffer, line 33: Just a reminder about this `TODO`.

Standard/src/Diagnostics/Internal.qs
```diff
@@ -0,0 +1,26 @@
...
+namespace Microsoft.Quantum.Diagnostics {
+    open Microsoft.Quantum.Intrinsic;
+
+    /// # Summary
+    /// Internal function used to fail with meaningful error messages.
+    ///
+    /// # Remarks
+    /// This function is intended to be emulated by simulation runtimes, so as
+    /// to allow forwarding formatted contradictions.
+    internal function FormattedFailure<'T>(actual : 'T, expected : 'T, message : String) : Unit {
```
tcNickolas, line 13: Why make this internal? I can see this being used in the external code as well.

Standard/tests/Diagnostics/AllowTests.qs
```diff
@@ -0,0 +1,57 @@
...
+                // Should run four times, exactly as
+                // many times as allowed.
+                for (idx in 0..3) {
+                    X(q);
+                }
+            }
+        }
+    }
+
+    operation CheckAllowNQubitsTestShouldFail() : Unit {
```
tcNickolas, line 37: I would add tests which would have "using - AllowAtMostNQubits - using" nesting (for both passing and failing scenario), to make sure it only tracks additional qubits allocated, per the documentation, and not all allocated qubits.

Standard/src/Diagnostics/Emulation/AllowOperationCalls.cs
```diff
@@ -0,0 +1,112 @@
...
+                            if (callSites.Count > nTimes)
+                            {
+                                Simulator?.MaybeDisplayDiagnostic(new CallSites
+                                {
+                                    Sites = callSites,
+                                    Subject = op.FullName
+                                });
+                                failed = true;
+                                throw new ExecutionFailException(
+                                    $"Operation {op.FullName} was called more than the allowed {nTimes} times."
```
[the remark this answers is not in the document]
cgranade: > If this exception upon calling `Adjoint AllowAtMostNCallsCA`, I'd suggest adding the number of times the operation was called to the message; it can be helpful to troubleshooting if you know by how much exactly you overuse the operation.

It fails on the (n +1)st call, as soon as the required condition fails; I agree, it may make sense to defer until the end of the allow-block, though.

Standard/src/Diagnostics/Emulation/AllowOperationCalls.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.Quantum.Simulation.Common;
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Quantum.Diagnostics
+{
+
```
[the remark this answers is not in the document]
cgranade: Agreed, there's no way to do exactly that using this feature. Given the short timescale until the next release, it may make sense to defer that as a feature request for September?

Visualization/src/Extensions.cs
cgranade: Thanks for catching that, @rmshaffer, I did indeed forget to update after  https://github.com/microsoft/iqsharp/pull/239 completed.

Standard/tests/Diagnostics/DumpTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Quantum.Simulation.Simulators;
+using Microsoft.Quantum.Standard.Emulation;
+using Xunit;
+using Assert = Xunit.Assert;
+
+namespace Microsoft.Quantum.Tests
+{
+    public class DumpOperation
+    {
+        [Fact]
```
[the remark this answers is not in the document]
cgranade: Good point, added that test.

Standard/src/Diagnostics/Dump.qs
```diff
@@ -0,0 +1,62 @@
...
+    /// the operation that are made available by the current
+    /// execution target.
+    ///
+    /// # Input
+    /// ## nQubits
+    /// The number of qubits on which the given operation acts.
+    /// ## op
+    /// The operation that is to be diagnosed.
+    ///
+    /// # Example
```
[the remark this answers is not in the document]
cgranade: I'm a little leery of promising a specific order in documentation, since it's dependent on internal conventions for each simulator that supports the operation; in particular, the `QuantumSimulator` class is free to rearrange how it orders diagnostics without breaking user code, insofar as all measurement probabilities are left invariant. I agree with the point, however, such that perhaps it might make sense to list the output from `QuantumSimulator` as an example, with an explicit note that conventions may vary.

Standard/src/Diagnostics/Internal.qs
cgranade: In general, my strategy has been to not increase API surface with utility functions and operations until we have a concrete usecase and have reviewed the API against design principles. Happy to revisit this case as part of a future feature request or PR, though.

cgranade:
> I made a pass and left some comments, mostly about documentation, tests and examples. Looks very promising!

Thanks for the feedback @tcNickolas, I appreciate it!

> We might be able to retire parts of [CounterSimulator](https://github.com/microsoft/QuantumKatas/blob/master/utilities/Common/CounterSimulator.cs), if not all of it.

Given your comments, it looks like there may be some usecases still left (e.g.: enforcing constraints on groups of operations), but yeah; my goal was to be able to consolidate that logic in a simulator-independent fashion, so that oracle counting policies could be applied against the Toffoli simulator, or even against third-party simulators.

> I'm assuming that DumpOperation works in command line Q# as well as in Jupyter notebooks, right?

That should be the case, yes. In particular, the `QuantumSimulator` class outputs the unitary representation of the given operation through the `SimulatorBase.OnDisplayableDiagnostic` event. Python and IQ# pick that up automatically and converts those diagnostics to either something Jupyter Notebook can read, or for command-line use. For C# host programs, adding a listener to that of the form `sim.OnDisplayableDiagnostic += Console.Out.WriteLine` should do it.

For Q# standalone programs, I'm not sure if that event is wired up automatically. It looks like https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/CsharpGeneration/EntryPoint.fs#L120 may need modified to add a listener to the `OnDisplayableDiagnostic` event. Alternatively, SimulatorBase.cs could be modified to add that listener by default and to disable it when `SimulatorBase.DisableLogToConsole` is called. Any thoughts, @bettinaheim?

In any case, I've added a `ToString` override `DisplayableUnitary` to make it easier to print the dumped diagnostics to the command line.

> I'd love to finally update [DumpUnitary](https://github.com/microsoft/QuantumKatas/tree/master/utilities/DumpUnitary) to dump the real unitary and not the column-by-column approximation!

That's one of the big advantages of using the Choi–Jamiołkowski isomorphism (also known in this case as AAPT) to find unitary representations; any relative phases between the various states in an input basis are directly captured by the use of a reference register, so that only phases that are truly global factor out. Still happy to share more about that if you're interested some time.

In the meantime, I've made a pass on your feedback and will update the PR accordingly. Thanks!

Standard/src/Diagnostics/Emulation/AllowQubitAllocations.cs
```diff
@@ -0,0 +1,85 @@
...
+
+                Handlers.Push((
+                    new ActionDisposer<Action<long>>(
+                        nQubits =>
+                        {
+                            nQubitsAllocated += nQubits;
+                            if (nQubitsAllocated > nQubitsAllowed)
+                            {
+                                throw new ExecutionFailException(
+                                    $"{nQubitsAllocated} were allocated, but at most {nQubitsAllowed} are allowed."
```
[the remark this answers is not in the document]
cgranade: Added message inputs to allow customizing failures.

Standard/tests/Diagnostics/AllowTests.qs
cgranade: Good suggestion, added said test.

Standard/src/Diagnostics/Dump.qs
```diff
@@ -34,6 +34,30 @@ namespace Microsoft.Quantum.Diagnostics {
...
+    /// \end{aligned}.
+    /// $$
+    ///
+    /// ```Q#
+    /// operation ApplyCnotToRegister(register : Qubit[]) : Unit is Adj + Ctl {
+    ///     CNOT(register[0], register[1]);
+    /// }
+    /// 
+    /// operation DumpCnot() : Unit {
+    ///     Diag.DumpOperation(2, ApplyCnotToRegister);
```
msoeken, line 57: How about `ApplyToFirstTwoQubitsCA(CNOT, _)` instead of `ApplyCnotToRegister`? Also should Cnot be CNOT in the operation names for consistency?

Standard/tests/Diagnostics/DumpTests.qs
```diff
@@ -10,4 +10,12 @@ namespace Microsoft.Quantum.Tests {
         Diag.DumpOperation(1, ApplyToEachCA(S, _));
     }
 
+    operation ApplyCnotToRegister(register : Qubit[]) : Unit is Adj + Ctl {
```
msoeken, line 13: See comments above

msoeken (approved):
Just left one small comment. Thanks

Standard/src/Diagnostics/Dump.qs
cgranade: Good point, was going for being explicit, but I agree with going for reuse. Thanks as well for the catch on acronyms and PascalCase interacting like that, I always mess that up.

vadym-kl:
Thank you @cgranade, this is very nice addition to the diagnostics namespace. It looks like there could be situations where `DumpOperation`  might have several dump formats. For example, in a Clifford simulator one might what to dump operation as a unitary matrix and also as an images of generators of n-qubit Pauli group. Do you have any thoughts on how to make `DumpOperation` a bit more feature proof for that?

cgranade:
> Thank you @cgranade, this is very nice addition to the diagnostics namespace. It looks like there could be situations where `DumpOperation` might have several dump formats. For example, in a Clifford simulator one might what to dump operation as a unitary matrix and also as an images of generators of n-qubit Pauli group. Do you have any thoughts on how to make `DumpOperation` a bit more feature proof for that?

Thanks for the kind words! Agreed, different simulators will likely have different formats for dumping operators; just as https://github.com/qsharp-community/chp-sim overrides DumpRegister and DumpMachine to show stabilizer tableaus, I'd love to see that project outputting Cliffords as tables of Pauli mappings, similar to what http://www.cgranade.com/python-quaec/ does. Similarly, I'd love to hook up ToffoliSimulator to output truth tables.

From that perspective, the design of this feature is intended to allow simulators defined downstream from Microsoft.Quantum.Standard to override DumpOperation, and for the DumpOperation emulation class add support for simulators defined upstream from Microsoft.Quantum.Standard.

PR #310, opened 2020-08-14
Add custom GetRuntimeMetadata for CX, CY, and CZ

theRoughCode:
This PR implements an override for the GetRuntimeMetadata methods for CX, CY, and CZ gates such that they match Controlled X, Y, and Z variants' RuntimeMetadata.

Without this, the IQ# ExecutionPathVisualizer would render CX, CY, and CZ as a 2-qubit unitary "box" gate rather than a controlled single-qubit unitary.


Standard/src/Canon/CommonGates.cs
```diff
@@ -0,0 +1,78 @@
...
+                return new RuntimeMetadata()
+                {
+                    Label = "X",
+                    IsControlled = true,
+                    Controls = new List<Qubit>() { ctrl },
+                    Targets = new List<Qubit>() { target },
+                };
+            }
+
+            return null;
```
cgranade, line 31: You can use C# 8 pattern matching to make this sort of logic a lot more compact.
```suggestion
            return args.Value switch
            {
                (Qubit ctrl, Qubit target) => new RuntimeMetadata()
                {
                    Label = "X",
                    IsControlled = true,
                    Controls = new List<Qubit>() { ctrl },
                    Targets = new List<Qubit>() { target },
                },
                _ => null
            };
```
Heck, you can even collapse further into an expression-body method by moving `Debug.Assert` call into the `_ => null` arm as a throw.

Standard/tests/Canon/CommonGatesTests.cs
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

Standard/tests/Canon/CommonGatesTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class CommonGatesRuntimeMetadataTests
+    {
+        [Fact]
+        public void CX()
+        {
+            var control = new MockQubit(1);
+            var target = new MockQubit(0);
+            var op = new Microsoft.Quantum.Canon.CX(new QuantumSimulator());
+            var args = op.__dataIn((control, target));
+            var expected = new RuntimeMetadata()
```
cgranade, line 28: Does `RuntimeMetadata` have a value equality comparison operator defined?

Standard/src/Canon/CommonGates.cs
theRoughCode: Wow, that is actually really nice. I wish I'd known that sooner as there are multiple places where I have this same structure. Love these C# pearls @cgranade!

Standard/tests/Canon/CommonGatesTests.cs
theRoughCode: Yup, in [here](https://github.com/microsoft/qsharp-runtime/blob/08ae4a499ea3f2c5c89ac8dbb0e4eb5b3c75c436/src/Simulation/Core/RuntimeMetadata.cs).

PR #318, opened 2020-08-31
Less-than comparison with a constant

msoeken:
This operation implements an optimised version of Fig. 17 in [arXiv:1902.02134](https://arxiv.org/abs/1902.02134v4). Constant input bits are propagated into the circuits, and no qubits are allocated for them. Further optimisations are applied for trailing 0s in the bit representation of the constant. The circuit uses AND gates instead of Toffoli gates to further reduce costs.

API review is necessary for consistent naming of arithmetic operations (e.g., compare to `CompareUsingRippleCarry` which performs greater-than comparison.)

### DIFF
--- a/Standard/src/Arithmetic/Arithmetic.qs
+++ b/Standard/src/Arithmetic/Arithmetic.qs
@@ -33,6 +33,31 @@ namespace Microsoft.Quantum.Arithmetic {
     }
 
     /// # Summary
+    /// Applies a bitwise-XOR operation between a classical integer and a
+    /// big integer represented by a register of qubits.
+    ///
+    /// # Description
+    /// Applies `X` operations to qubits in a little-endian register based on
+    /// 1 bits in a big integer.
+    ///
+    /// Let us denote `value` by a and let y be an unsigned integer encoded in `target`,
+    /// then `InPlaceXorLE` performs an operation given by the following map:
+    /// $\ket{y}\rightarrow \ket{y\oplus a}$ , where $\oplus$ is the bitwise exclusive OR operator.
+    ///
+    /// # Input
+    /// ## value
+    /// A big integer which is assumed to be non-negative.
+    /// ## target
+    /// A quantum register which is used to store `value` in little-endian encoding.
+    operation ApplyXorInPlaceL(value : BigInt, target : LittleEndian)
+    : Unit is Adj + Ctl {
+        ApplyToEachCA(
+            CControlledCA(X),
+            Zip(BigIntAsBoolArray(value), target!)
+        );
+    }
+
+    /// # Summary
     /// Applies the three-qubit majority operation in-place on a register of
     /// qubits.
     ///
--- a/Standard/src/Arithmetic/Comparators.qs
+++ b/Standard/src/Arithmetic/Comparators.qs
@@ -73,6 +73,7 @@ namespace Microsoft.Quantum.Arithmetic {
         } else {
             let l = TrailingZeroes(c);
             using ((tmpConstants, tmpAnd) = (Qubit[bitwidth - l], Qubit[bitwidth - 2 - l])) {
+                let tmpCarry = x![l..l] + tmpAnd;
                 within {
                     ApplyXorInPlaceL(c >>> l, LittleEndian(tmpConstants));
                     ApplyXorInPlaceL(c, x);
@@ -80,18 +81,18 @@ namespace Microsoft.Quantum.Arithmetic {
                         if (i == 0) {
                             CNOT(x![i + l], tmpConstants[i + 1]);
                         } else {
-                            ApplyAnd(tmpConstants[i], x![i + l], tmpAnd[i - 1]);
+                            ApplyAnd(tmpConstants[i], x![i + l], tmpCarry[i]);
                             if (i == 1) {
-                                CNOT(x![i + l - 1], tmpAnd[i - 1]);
+                                CNOT(x![i + l - 1], tmpCarry[i]);
                             } else {
-                                CNOT(tmpAnd[i - 2], tmpAnd[i - 1]);
+                                CNOT(tmpCarry[i - 1], tmpCarry[i]);
                             }
-                            CNOT(tmpAnd[i - 1], tmpConstants[i + 1]);
+                            CNOT(tmpCarry[i], tmpConstants[i + 1]);
                         }
                     }
                 } apply {
                     ApplyAnd(Tail(tmpConstants), Tail(x!), output);
-                    CNOT(Length(tmpAnd) == 0 ? x![l] | Tail(tmpAnd), output);
+                    CNOT(Tail(tmpCarry), output);
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


Standard/src/Arithmetic/Comparators.qs
```diff
@@ -56,4 +59,95 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// - Qubitization of Arbitrary Basis Quantum Chemistry Leveraging Sparsity and Low Rank Factorization
+    ///   Dominic W. Berry, Craig Gidney, Mario Motta, Jarrod R. McClean, Ryan Babbush
+    ///   Quantum 3, 208 (2019)
+    ///   https://arxiv.org/abs/1902.02134v4
+    operation LessThanConstantUsingRippleCarry(c : BigInt, x : LittleEndian, output : Qubit)
+    : Unit {
+        body (...) {
+            let bitwidth = Length(x!);
+            AssertAllZero([output]);
+            Fact(c >= 0L, "Constant input `c` must not be negative");
```
cgranade, line 100:
```suggestion
            Fact(c >= 0L, $"Constant input `c` must not be negative, but was {c}");
```

Standard/src/Arithmetic/Comparators.qs
```diff
@@ -56,4 +59,95 @@ namespace Microsoft.Quantum.Arithmetic {
...
+            if (c == 0L) {
+                // do nothing; output stays 0
+            } elif (c >= (1L <<< bitwidth)) {
+                X(output);
+            } elif (c == (1L <<< (bitwidth - 1))) {
+                CNOT(Tail(x!), output);
+                X(output);
+            } else {
+                let bits = BigIntAsBoolArray(c);
+                let l = IndexOf(EqualB(true, _), bits);
```
cgranade, line 111:
```suggestion
                let l = IndexOf(Identity<Bool>, bits);
```

Standard/src/Arithmetic/Comparators.qs
```diff
@@ -56,4 +59,95 @@ namespace Microsoft.Quantum.Arithmetic {
...
+                    let tmpCarry = x![l..l] + tmpAnd;
+                    within {
+                        ApplyPauliFromBitString(PauliX, true, bits, x!);
+                        for (i in 1..bitwidth - l - 2) {
+                            within {
+                                ApplyIfA(X, bits[i + l], tmpCarry[i - 1]);
+                            } apply {
+                                ApplyAnd(tmpCarry[i - 1], x![i + l], tmpCarry[i]);
+                            }
+                            CNOT(tmpCarry[i - 1], tmpCarry[i]);
```
cgranade, line 122: This is precisely in common with the block at 125 to 130, up to the particular inputs being passed; would it be good to refactor into an internal operation?

Standard/src/Arithmetic/Comparators.qs
```diff
@@ -56,4 +59,95 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// is 1, after the trailing `0`s have been removed.  Further, all qubits associated
+    /// to constant inputs in the construction of the original work are propagated and
+    /// not allocated in this implementation.
+    ///
+    /// # References
+    /// - Qubitization of Arbitrary Basis Quantum Chemistry Leveraging Sparsity and Low Rank Factorization
+    ///   Dominic W. Berry, Craig Gidney, Mario Motta, Jarrod R. McClean, Ryan Babbush
+    ///   Quantum 3, 208 (2019)
+    ///   https://arxiv.org/abs/1902.02134v4
+    operation LessThanConstantUsingRippleCarry(c : BigInt, x : LittleEndian, output : Qubit)
```
cgranade, line 95: We can defer to API review, but I'd suggest maybe something like "CompareLessThanConstantUsingRippleCarry"?

Standard/src/Arithmetic/Comparators.qs
```diff
@@ -56,4 +59,95 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    : Unit {
```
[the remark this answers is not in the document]
msoeken: Good point, I wasn't aware that this is possible.

Standard/src/Arithmetic/Comparators.qs
msoeken: I didn't know this one! Couldn't find it in the docs though.

Standard/src/Arithmetic/Arithmetic.qs
```diff
@@ -32,6 +32,31 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// Let us denote `value` by a and let y be an unsigned integer encoded in `target`,
+    /// then `InPlaceXorLE` performs an operation given by the following map:
+    /// $\ket{y}\rightarrow \ket{y\oplus a}$ , where $\oplus$ is the bitwise exclusive OR operator.
+    ///
+    /// # Input
+    /// ## value
+    /// A big integer which is assumed to be non-negative.
+    /// ## target
+    /// A quantum register which is used to store `value` in little-endian encoding.
+    operation ApplyXorInPlaceL(value : BigInt, target : LittleEndian)
```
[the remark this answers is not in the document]
msoeken: Added one and also another one to `ApplyXorInPlace`

Standard/src/Arithmetic/Comparators.qs
msoeken: Sounds good, but we need further discussion because we also have `CompareUsingRippleCarry` at the moment.

Standard/tests/ArithmeticTests.qs
```diff
@@ -173,7 +176,48 @@ namespace Microsoft.Quantum.ArithmeticTests {
...
+    }
+
+    @Test("ResourcesEstimator")
+    operation TestLessThanConstantUsingRippleCarryOperationCalls() : Unit {
+        let tCounts = [0, 12, 8, 12, 4, 12, 8, 12, 0, 12, 8, 12, 4, 12, 8, 12, 0];
+        let qubitCounts = [0, 2, 1, 2, 0, 2, 1, 2, 0, 2, 1, 2, 0, 2, 1, 2, 0];
+
+        for ((idx, (tCount, qubitCount)) in Enumerated(Zip(tCounts, qubitCounts))) {
+            within {
+                AllowAtMostNCallsCA(tCount * 4, T, $"Too many T operations for constant {idx}");
```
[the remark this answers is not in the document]
msoeken: Yes, I think that `AllowExactlyNCallsCA` could be useful as well. Because I know the exact counts here.

Standard/src/Arithmetic/Comparators.qs
msoeken: Yes, did that, and found a way to save that CNOT gate by doing so.

msoeken:
Thanks for your feedback @cgranade, I have incorporated your comments.

Standard/src/Arithmetic/Comparators.qs
cgranade: Good point; that can definitely create some confusion. Agreed with the need for further discussion.

Issue #313, opened 2020-08-21
Improvements to Microsoft.Quantum.Arrays


cgranade:
# Abstract
There is a close connection between the iteration over classical values, and generalizing "circuit"-like patterns into high-level code. In Q#, we use exploit this connection with the Microsoft.Quantum.Arrays namespace to make it easier express high-level patterns. We have noted from user feedback, however, that some common patterns are difficult to express using existing array functions, or that users may not be able to discover needed functionality.

The improvements proposed here empower users to make better use of high-level quantum development in Q# by making it easier to iterate over collections of values, such as Qubit[], and to use them to write out complex algorithmic patterns efficiently.

# Proposed deprecations and renamings

- namespace Microsoft.Quantum.Arrays 
  - <s>`function Exclude<'T>(remove : Int[], array : 'T[]) : 'T[];`</s>
    `function Excluding<'T>(remove : Int[], array : 'T[]) : 'T[]; `
  - <s>`function Zip<'T, 'U>(left : 'T[], right : 'U[]) : ('T, 'U)[]; `</s>
    `function Zipped<'T, 'U>(left : 'T[], right : 'U[]) : ('T, 'U)[]; `

# Proposed additions

- namespace Microsoft.Quantum.Arrays 
  - Accessor functions 
    - [`function ElementAt<'T>(index : Int, array : 'T[]) : 'T; `](https://github.com/microsoft/QuantumLibraries/pull/320)
    - [`function ElementsAt<'T>(range : Range, array : 'T[]) : 'T[];`](https://github.com/microsoft/QuantumLibraries/pull/320)
  - Reductions and combinators 
    - [`function CumulativeFolded<'State, 'TInput>(fn : ('State, 'TInput -> 'State), initialState : 'State, arr : 'TInput[]) : 'State[];`](https://github.com/microsoft/QuantumLibraries/pull/320)
    - [`function FlatMapped<'TInput, 'TOutput>(mapper : 'TInput -> 'TOutput[], arr : 'TInput[]) : 'TOutput[];`](https://github.com/microsoft/QuantumLibraries/pull/320)
      Returns (mapper(arr[0]) + mapper(arr[1]) + ... + mapper(arr[n - 1])). 
    - [`function MappedOverRange<'T>(fn : Int -> 'T, range : Range) : 'T[];`](https://github.com/microsoft/QuantumLibraries/pull/320)
      Returns the value of `fn` at each element of `range`.
[...]

PR #320, opened 2020-09-01
New array functions

msoeken:
This PR proposed some new functions for the `M.Q.Arrays` namespace (PR into feature branch).


Standard/src/Arrays/Zip.qs
```diff
@@ -131,6 +131,40 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// # Output
+    /// Two arrays, the first one containing all first elements of the input
+    /// tuples, the second one containing all second elements of the input tuples.
+    ///
+    /// # Example
+    /// ```Q#
+    /// // split is same as ([6, 5, 5, 3, 2, 1], [true, false, false, false, true, false])
+    /// let split = Unzipped([(6, true), (5, false), (5, false), (3, false), (2, true), (1, false)]);
+    /// ```
+    function Unzipped<'T, 'U>(arr : ('T, 'U)[]) : ('T[], 'U[]) {
```
cgranade, line 157: A `/// # See Also` back to `Zip` may also be helpful here.

Standard/tests/ArrayTests.qs
```diff
@@ -26,6 +26,16 @@ namespace Microsoft.Quantum.Tests {
         }
     }
 
+    @Test("QuantumSimulator")
+    function UnzippedTest() : Unit {
```
cgranade, line 30: Currently, due to the C# test discoverer being used alongside the Q# unit testing attributes, having a function or operation whose name ends with `Test` in the `Microsoft.Quantum.Tests` namespace and is annotated with `@Test` will result in it being run twice. That's definitely a pre-existing issue, and is likely not a huge problem for a test as simple as this one, but it's still good to keep in mind, and can be avoided by changing the namespace name (e.g.: to `Microsoft.Quantum.Arrays.Tests`).

cgranade:
Added to https://github.com/microsoft/QuantumLibraries/issues/313 for review.

msoeken:
Thanks for the feedback @cgranade. I made a revision.

Standard/tests/ArrayTests.qs
msoeken: Thanks. I also changed some others in the same file.

cgranade:
> Thanks for the feedback @cgranade. I made a revision.

Looks good to me; will defer formal approval until #313 passes API review, but I think this should be fine. Thanks!

msoeken:
Thanks @cgranade. This PR is still marked as draft. I wanted to add some more array functions before merging it into the feature branch.

Standard/src/Arrays/Windows.qs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Arrays {
+    /// # Summary
```
cgranade, line 5:
```suggestion
namespace Microsoft.Quantum.Arrays {

    /// # Summary
```

Standard/src/Arrays/Zip.qs
```diff
@@ -131,6 +132,43 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// # Output
+    /// Two arrays, the first one containing all first elements of the input
+    /// tuples, the second one containing all second elements of the input tuples.
+    ///
+    /// # Example
+    /// ```Q#
+    /// // split is same as ([6, 5, 5, 3, 2, 1], [true, false, false, false, true, false])
+    /// let split = Unzipped([(6, true), (5, false), (5, false), (3, false), (2, true), (1, false)]);
+    /// ```
+    ///
```
cgranade, line 158: Would it help to add a remark that this is equivalent to `(Mapped(Fst<'T, 'U>, arr), Mapped(Snd<'T, 'U>, arr))`?

cgranade (commented):
Looks good so far, mainly had some comments about API documentation comments.

Standard/src/Arrays/Interleaved.qs
```diff
@@ -0,0 +1,44 @@
...
+    /// to pick the first element from the first array, then the first
+    /// element from the second array, and so on.  The first array is either
+    /// of the same length as the second one, or can have one more element.
+    ///
+    /// # Input
+    /// ## first
+    /// First array
+    ///
+    /// ## second
+    /// Second array
```
msoeken, line 21:
```suggestion
    /// Second array to be interleaved.
```

Standard/src/Arrays/Multidimensional.qs
```diff
@@ -0,0 +1,133 @@
...
+namespace Microsoft.Quantum.Arrays {
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Convert;
+    open Microsoft.Quantum.Diagnostics;
+
+    /// # Summary
+    /// Returns the transpose of a matrix
+    ///
+    /// # Description
+    /// Input as an $r \times c$ matrix with $r$ rows and $$ columns.  The matrix
```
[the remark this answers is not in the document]
msoeken: Good catch! Thanks

Standard/src/Arrays/Multidimensional.qs
```diff
@@ -0,0 +1,133 @@
...
+    /// # Example
+    /// ```Q#
+    /// // same as [[1, 4], [2, 5], [3, 6]]
+    /// let transposed = Transposed([[1, 2, 3], [4, 5, 6]]);
+    /// ```
+    function Transposed<'T>(matrix : 'T[][]) : 'T[][] {
+        let numRows = Length(matrix);
+        Fact(numRows > 0, "Matrix must have at least 1 row");
+        let numColumns = Length(Head(matrix));
+        Fact(numColumns > 0, "Matrix must have at least 1 column");
```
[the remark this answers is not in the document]
msoeken: Yes, I'll add diagnostic functions for this case and then use them here.

Standard/src/Arrays/Multidimensional.qs
```diff
@@ -0,0 +1,133 @@
...
+    /// Returns the array's element at given index
+    ///
+    /// # Input
+    /// ## index
+    /// Index of element
+    /// ## array
+    /// Array
+    ///
+    /// # Remark
+    /// This function is more general than `LookupFunction`, since
```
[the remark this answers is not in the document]
msoeken: That would make sense. Would we lose performance for additional nesting?

Standard/src/Arrays/Multidimensional.qs
```diff
@@ -0,0 +1,133 @@
...
+    /// - Microsoft.Quantum.Arrays.ElementAt
+    /// - Microsoft.Quantum.Arrays.LookupFunction
+    function ElementsAt<'T>(range : Range, array : 'T[]) : 'T[] {
+        return array[range];
+    }
+
+    /// # Summary
+    /// Extracts a column in a matrix
+    ///
+    /// # Dimension
```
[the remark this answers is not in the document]
msoeken: I meant to say `Description` here, but my head must have been in another dimension 😄

Standard/src/Arrays/Zip.qs
msoeken: Yes, I can add this. In fact this was my first implementation for this function, but it did want to avoid iterating through the array twice.

cgranade: Yeah, I definitely like the way you currently have it better; just seems like it can be helpful as a comment to establish how the different array functions work together. Thanks!

Standard/src/Arrays/Multidimensional.qs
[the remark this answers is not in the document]
cgranade: Possibly, yes. That would be good to validate, but I would also expect the performance penalty to be quite small, since `LookupFunction` already works by partial application over a very similar internal function.

Standard/src/Arrays/Reductions.qs
```diff
@@ -0,0 +1,55 @@
...
+    /// Combines Mapped and Fold into a single function
+    ///
+    /// # Description
+    /// This function iterates the `folder` function through the array, starting from
+    /// an initial state `state` and returns all intermediate values, not including
+    /// the inital state.
+    ///
+    /// # Type Parameters
+    /// ## 'State
+    /// The type of states the `folder` function operates on, i.e., accepts as its first
```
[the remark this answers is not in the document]
cgranade: Ah, got it. That makes more sense!

Standard/src/Arrays/Windows.qs
```diff
@@ -0,0 +1,76 @@
...
+    /// The type of `array` elements.
+    ///
+    /// # Input
+    /// ## array
+    /// An array of elements.
+    ///
+    /// # Example
+    /// ```Q#
+    /// let prefixes = Prefixes([23, 42, 144]);
+    /// // prefixes = [EmptyArray<Int>(), [23], [23, 42], [23, 42, 144]]
```
cgranade, line 67: It may be worth considering leaving out the empty prefix for consistency with CumulativeFolded; as with that case, I think either is fine, but it may be confusing to use different conventions in both cases without a clear and predictable reason to do so.

Standard/tests/ArrayTests.qs
```diff
@@ -237,6 +256,86 @@ namespace Microsoft.Quantum.Tests {
...
+        EqualityFactI(Length(prefixes), 7, "unexpected length for prefixes");
+        EqualityFactI(Length(prefixes[0]), 0, "unexpected prefix");
+        AllEqualityFactI(prefixes[1], [0], "unexpected prefix");
+        AllEqualityFactI(prefixes[2], [0, 1], "unexpected prefix");
+        AllEqualityFactI(prefixes[3], [0, 1, 1], "unexpected prefix");
+        AllEqualityFactI(prefixes[4], [0, 1, 1, 2], "unexpected prefix");
+        AllEqualityFactI(prefixes[5], [0, 1, 1, 2, 3], "unexpected prefix");
+        AllEqualityFactI(prefixes[6], [0, 1, 1, 2, 3, 5], "unexpected prefix");
+    }
 }
```
cgranade, line 339: Would it make sense to test the square and rectangular matrix facts themselves? You may need to use the legacy testing framework to do so, ending test callables with "TestShouldFail."

cgranade (commented):
I think this is nearly there, thanks for doing so much to implement all this!

Standard/src/Arrays/Multidimensional.qs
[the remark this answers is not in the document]
msoeken: I refactored this. The internal function could then be removed.

Standard/src/Arrays/Multidimensional.qs
```diff
@@ -0,0 +1,268 @@
...
+    /// ```Q#
+    /// let matrix = [[1, 2, 3], [4, 5, 6], [7, 8, 9]];
+    /// let diagonal = Diagonal(matrix);
+    /// // same as: column = [1, 5, 9]
+    /// ```
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Arrays.Transposed
+    function Diagonal<'T>(matrix : 'T[][]) : 'T[] {
+        SquareArrayFact(matrix, "Matrix is not a rectangular array");
```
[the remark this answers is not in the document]
msoeken: Yes, I will change that.

Standard/src/Arrays/Reductions.qs
```diff
@@ -0,0 +1,55 @@
...
+    /// Combines Mapped and Fold into a single function
+    ///
+    /// # Description
+    /// This function iterates the `folder` function through the array, starting from
+    /// an initial state `state` and returns all intermediate values, not including
+    /// the inital state.
+    ///
+    /// # Type Parameters
+    /// ## 'State
+    /// The type of states that the `folder` function operates on, i.e., accepts as its first
```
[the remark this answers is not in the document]
msoeken: Will change in next commit.

Standard/src/Arrays/Windows.qs
msoeken: Yes, that makes sense. I will also update the issue description.

cgranade (approved):

Issue #317, opened 2020-08-28
Use of "__" in Q# identifiers will be reserved in the September QDK release


samarsha:
The following operation names contain `__`:

* `Microsoft.Quantum.Chemistry.JordanWigner._ToJordanWignerSelectInput__`
* `Microsoft.Quantum.Chemistry.JordanWigner._ToPauliBases__`
* `Microsoft.Quantum.Chemistry.JordanWigner._JordanWignerOptimizedBlockEncodingStatePrep__`
* `Microsoft.Quantum.Chemistry.JordanWigner._OptimizedBEXY__`
* `Microsoft.Quantum.Chemistry.JordanWigner._SelectZ__`
* `Microsoft.Quantum.Chemistry.JordanWigner._JordanWignerSelect__`

This will be reserved in the September QDK release, and should be renamed. (See microsoft/qsharp-compiler#578.)

PR #323, opened 2020-09-08
Fix several warnings

msoeken:
Fixes several warnings in Standard.sln and addresses #317.

### DIFF
--- a/Standard/src/Characterization/EstimateFrequency.cs
+++ b/Standard/src/Characterization/EstimateFrequency.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Runtime.ExceptionServices;
@@ -70,7 +73,7 @@ namespace Microsoft.Quantum.Characterization
                 if (paulis.Length != count) throw new InvalidOperationException("The number of paulis must match the number of qubits.");
 
                 var qubits = this.Allocate.Apply(count);
-                Exception? innerException = null;
+                Exception innerException = null;
                 double result = 0.0;
                 try
                 {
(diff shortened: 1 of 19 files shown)


Standard/src/Characterization/EstimateFrequency.cs
```diff
@@ -70,7 +73,7 @@ public override void __Init__()
                 if (paulis.Length != count) throw new InvalidOperationException("The number of paulis must match the number of qubits.");
 
                 var qubits = this.Allocate.Apply(count);
-                Exception? innerException = null;
+                Exception innerException = null;
```
cgranade, line 76: This looks a bit odd to me, in that the type on the left is a non-nullable reference type, but the value on the right is `null`. Perhaps the issue is that `#nullable enable` doesn't seem to be present in the preamble for this file?

Standard/src/Diagnostics/Emulation/DataStructures.cs
```diff
@@ -40,7 +40,7 @@ public class DisplayableUnitaryOperator
         public override string ToString() =>
             Data == null
             ? ""
-            : $"Real:\n{Data[Ellipsis, 0]}\nImag:\n{Data[Ellipsis, 1]}";
+            : $"Real:\n{Data![Ellipsis, 0]}\nImag:\n{Data![Ellipsis, 1]}";
```
cgranade, line 43: Interesting, the compiler should be able to figure out that `Data` isn't null in this branch of the `?:` operator; perhaps this will be fixed with general improvements to the `?:` operator in C# 9...!

cgranade (approved):
Looks good to me, thanks for doing this! Just had the one question about nullability in a C# fix.

Standard/src/Characterization/EstimateFrequency.cs
msoeken: Yes, there's no `#nullable enable` in the header. I tried adding it, but it added a lot of other warnings and it seems there was no trivial fix, so I suggest to address it in a separate PR.

Standard/src/Diagnostics/Emulation/DataStructures.cs
msoeken: I also thought so and was surprised to retrieve the warning. The same for another location in which `Data` was checked for `null` and an exception is thrown in that case.

PR #324, opened 2020-09-08
Improved assertion for PrepareUniformSuperposition

### LINKED ISSUE #221: PrepareUniformSuperposition is not generally adjointable
**Describe the bug**
In the signature for `PrepareUniformSuperposition`, it says it should be adjointable, but it seems like it only is adjointable if you start with the uniform superposition state (ie. do `PrepareUniformSuperposition` then `Adjoint PrepareUniformSuperposition`)

**To Reproduce**
- This snippet crashes at runtime:
```qsharp
open Microsoft.Quantum.Preparation;
open Microsoft.Quantum.Arithmetic;

operation Foo() : Unit {
    using (qs = Qubit[4]) {
        Adjoint PrepareUniformSuperposition(10, LittleEndian(qs));
        DumpMachine();
        PrepareUniformSuperposition(10, LittleEndian(qs));
        ResetAll(qs);
    }
}
```

- If you reverse the order of the adjoint in the snippet: 

```qsharp
open Microsoft.Quantum.Preparation;
open Microsoft.Quantum.Arithmetic;

operation Bar() : Unit {
    using (qs = Qubit[4]) {
        PrepareUniformSuperposition(10, LittleEndian(qs));
        DumpMachine();
        Adjoint PrepareUniformSuperposition(10, LittleEndian(qs));
        ResetAll(qs);
    }
}
```

**Expected behavior**
I would expect that it would not crash when I try to run Adjoint of the operation and then the operation which should do the same as if you do the operation and then its Adjoint.

**Screenshots**
Here is running of the sample on the binder instance at [aka.ms/try-qsharp](aka.ms/try-qsharp)
image

**System information**
iqsharp	0.10.1912.501
Jupyter Core	1.2.20112.0
.NET Runtime	.NETCoreApp,Version=v3.0

msoeken:
This is a proposal to fix #221. It adds an assertion that the input state is in state `|0...0>`. The assertion will also get triggered when calling the `Adjoint` variant on a state that is not in a uniform superposition. Ideally, we would check whether the state is eventually in a uniform superposition, but this may add too much overhead.

### DIFF
--- a/Standard/src/Preparation/UniformSuperposition.qs
+++ b/Standard/src/Preparation/UniformSuperposition.qs
@@ -12,7 +12,7 @@ namespace Microsoft.Quantum.Preparation {
     open Microsoft.Quantum.Math;
 
     /// # Summary
-    /// Creates a uniform superposition over states that encode 0 through `nIndices`.
+    /// Creates a uniform superposition over states that encode 0 through `nIndices - 1`.
     ///
     /// That is, this unitary $U$ creates a uniform superposition over all number states
     /// $0$ to $M-1$, given an input state $\ket{0\cdots 0}$. In other words,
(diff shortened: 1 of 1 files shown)


cgranade:
I think this makes a lot of sense, though I'd also suggest adding an explicit note to the API documentation comment, since its not obvious _a priori_ that the operation is adjointable only as a partial isometry and not as a unitary.

cgranade (approved):
Approved, modulo comment about documentation. Thanks!

PR #325, opened 2020-09-14
End-to-end build

anpaz-msft:
Includes updating references from `master` to `main`


Build/e2e.yml
```diff
@@ -0,0 +1,33 @@
...
+
+parameters:
+- name: validation_level
+  displayName: Validation Level
+  type: string
+  default: normal
+  values:
+  - minimal
+  - normal
+  - full
```
cgranade, line 11: Oh, that's neat. I didn't realize you could make enums now!

Build/e2e.yml
```diff
@@ -0,0 +1,33 @@
...
+
+trigger: none
+
+pr:
+- main
+- feature/*
```
cgranade, line 17: Does this result in a warning or error for feature branches? I could see cases where e2e legitimately fails for feature branches still under development.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.1.cs
```diff
@@ -128,7 +128,7 @@ internal static class BroombridgeExtensionsV0_1
     #region Broombridge v0.1 format
     public static class V0_1
     {
-        public static string SchemaUrl = "https://raw.githubusercontent.com/microsoft/Quantum/master/Chemistry/Schema/broombridge-0.1.schema.json";
```
cgranade, line 131: I think I mentioned in the other PR as well, so apologies if this duplicates, but my only concern here would be making sure that we don't break old Broombridge files.

cgranade (approved):

Build/e2e.yml
anpaz-msft: It will trigger a build, and if it fails the check will report the failure but because we're not making it "Required", you can still commit. We will make it Required only for `main`, but that is a setting in the repo, not in the yaml.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.1.cs
anpaz-msft: I'm hoping that automatic redirects would take care of that, but apparently it doesn't work.
The only option is to keep the old `master` branch around, although stale, to avoid failures.

PR #331, opened 2020-09-22
Combinators for ApplyTo of an operation to Head, Tail, Rest, and Most of an array

msoeken:
This implements the `ApplyTo{Head,Rest,Tail,Most}{,C,A,CA}` combinators described in #313.

### DIFF
--- a/Standard/src/Canon/Combinators/ApplyToArray.qs
+++ b/Standard/src/Canon/Combinators/ApplyToArray.qs
@@ -1,0 +1,406 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Canon {
+    open Microsoft.Quantum.Arrays;
+
+    /// # Summary
+    /// Applies an operation to the first element of an array.
+    ///
+    /// # Description
+    /// Given an operation `op` and an array of targets `targets`,
+    /// applies `op(Head(targets))`.
+    ///
+    /// # Input
+    /// ## op
+    /// An operation to be applied.
+    /// ## target
+    /// An array of targets, of which the first will be applied to `op`.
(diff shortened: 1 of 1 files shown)


Standard/src/Canon/Combinators/ApplyToArray.qs
```diff
@@ -0,0 +1,406 @@
...
+    ///
+    /// # Type Parameters
+    /// ## 'T
+    /// The input type of the operation to be applied.
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.ApplyToHeadA
+    /// - Microsoft.Quantum.Canon.ApplyToHeadC
+    /// - Microsoft.Quantum.Canon.ApplyToHeadCA
+    operation ApplyToHead<'T>(op : ('T => Unit), targets : 'T[]) : Unit {
```
cgranade, line 28: I'd suggest that some examples would be helpful here, such as the following:
```suggestion
    ///
    /// # Example
    /// The following Q# snippets are equivalent:
    /// ```Q#
    /// ApplyToHead(H, register);
    /// H(Head(register));
    /// ```
    operation ApplyToHead<'T>(op : ('T => Unit), targets : 'T[]) : Unit {
```

cgranade (commented):
I think this looks good so far, but I'd suggest additional unit tests and API documentation examples before merging this in.

msoeken:
Thanks for your comments @cgranade. I have added an example to each of the *main* operations (the one without `A`, `C`, and `CA` modifiers).

guenp (approved):
Looks good to me! Just out of curiosity, not sure how often this use case would come up but would it also be useful to apply an operation to any given element `N` of an array or all even/odd elements?

PR #332, opened 2020-09-23
Array function to filter for unique elements

msoeken:
This implements `Unique` from the proposal in #313.

Note that this function implements `Unique` wrt. the semantics, e.g., found in C++.

It returns an array with the same order but no adjacent equal elements. If one wants to have _truly_ unique arrays, one can combine this function with `Sorted` (mentioned in the docs).

If we want a different semantics, we might need to add a comparison function to check for order, or a more general comparison function that returns (-1 for `<`, 0 for `=`, and 1 for `>`).

### DIFF
--- a/Standard/src/Arrays/Unique.qs
+++ b/Standard/src/Arrays/Unique.qs
@@ -1,0 +1,59 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Arrays {
+    /// # Summary
+    /// Returns a new array that has no equal adjacent elements.
+    ///
+    /// # Description
+    /// Given some array of elements and a function to test equality, this
+    /// function returns a new array in which the relative order of elements
+    /// is kept, but all adjacent elements which are equal are filtered to
+    /// just a single element.
+    ///
+    /// # Type Parameters
+    /// ## 'T
+    /// The type of each element of `array`.
+    ///
+    /// # Input
+    /// ## equal
+    /// A function that compares two elements such that `a` is considered to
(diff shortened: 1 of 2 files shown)


cgranade (approved):
Looks good, thanks for adding this!

Standard/src/Arrays/Unique.qs
```diff
@@ -0,0 +1,59 @@
...
+    /// # Example
+    /// ```Q#
+    /// let unique1 = Unique(EqualI, [1, 1, 3, 3, 2, 5, 5, 5, 7]);
+    /// // same as [1, 3, 2, 5, 7]
+    /// let unique2 = Unique(EqualI, [2, 2, 1, 1, 2, 2, 1, 1]);
+    /// // same as [2, 1, 2, 1];
+    /// let unique3 = Unique(EqualI, Sorted(LessThanOrEqualI, [2, 2, 1, 1, 2, 2, 1, 1]));
+    /// // same as [1, 2];
+    /// ```
+    function Unique<'T>(equal : (('T, 'T) -> Bool), array : 'T[]) : 'T[] {
```
guenp, line 42: I think the name `Unique` suggests that this would only return unique elements and I can see this becoming confusing for users (at least myself would be confused). Since it only promises unique _adjacent_ elements, would it be an idea to name this `UniqueAdjacent` or `DroppedDuplicates` instead?

guenp (changes requested):
Looks like a good and useful addition! Just would suggest different naming. And, would it make sense to actually add a `Unique` function as well, that returns all unique elements in an array?

guenp (approved):
After some discussion, it looks like this was already discussed in the API design meeting, so I approve this PR.

PR #333, opened 2020-09-23
Deprecate Zip, Exclude

cgranade:
This PR takes the action described in e72b1e9 and deprecates `Zip` for `Zipped, and `Exclude` for `Excluding`. As per that discussion, we should also consider how to help users discover new names.

### DIFF
--- a/Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
+++ b/Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
@@ -10,7 +10,7 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
     open Microsoft.Quantum.Arrays;
 
     /// # Summary
-    /// Computes Z component of Jordan–Wigner string between
+    /// Computes Z component of Jordanï¿½Wigner string between
     /// fermion indices in a fermionic operator with an even
     /// number of creation / annihilation operators.
     ///
@@ -89,7 +89,7 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
         let ops = [[PauliX, PauliY], [PauliY, PauliX]];
         let signs = [+1.0, -1.0];
 
-        for ((op, sign) in Zip(ops, signs)) {
+        for ((op, sign) in Zipped(ops, signs)) {
             let pauliString = _ComputeJordanWignerPauliString(Length(qubits), idxFermions, op);
             Exp(pauliString, sign * angle, qubits);
         }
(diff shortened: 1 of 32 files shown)


Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
     open Microsoft.Quantum.Arrays;
 
     /// # Summary
-    /// Computes Z component of Jordan�Wigner string between
+    /// Computes Z component of Jordan�Wigner string between
```
guenp, line 13: what is this � character?

MachineLearning/src/Structure.qs
```diff
@@ -63,7 +63,7 @@ namespace Microsoft.Quantum.MachineLearning {
     }
 
     function _UncontrolledSpanSequence(idxsQubits : Int[]) : (Int, Int[])[] {
-        return Zip(
+        return ped(
```
guenp, line 66:
```suggestion
        return Zipped(
```

guenp (changes requested):

Chemistry/src/Runtime/JordanWigner/JordanWignerClusterOperatorEvolutionSet.qs
cgranade: A bug... thanks for catching that. Let me go fix it.
```suggestion
    /// Computes Z component of Jordan–Wigner string between
```

Chemistry/src/Runtime/JordanWigner/JordanWignerEvolutionSet.qs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Quantum.Chemistry.JordanWigner {
     open Microsoft.Quantum.Chemistry;
     open Microsoft.Quantum.Arrays;
 
-    // This evolution set runs off data optimized for a Jordan�Wigner encoding.
+    // This evolution set runs off data optimized for a Jordan�Wigner encoding.
```
guenp, line 11:
```suggestion
    // This evolution set runs off data optimized for a Jordan-Wigner encoding.
```

cgranade: If it's alright, I'd like to make a slightly different correction (I get a bit picky about hyphens versus en dashes for separating names).
```suggestion
    // This evolution set runs off data optimized for a Jordan–Wigner encoding.
```

guenp (approved):

PR #342, opened 2020-10-07
Add metadata needed for improved documentation generation

cgranade:
This PR adds package names to documentation metadata, adapts C# magic command implementations to resolve the ambiguity between `Documentation` from the Microsoft.Quantum.* and Microsoft.Jupyter.Core namespaces, and adds `/// # Named Items` sections. See https://github.com/microsoft/qsharp-compiler/pull/594 for additional context and discussion.


Numerics/src/Numerics.csproj
```diff
@@ -4,12 +4,13 @@
...
     <AssemblyName>Microsoft.Quantum.Numerics</AssemblyName>
     <QsharpDocsGeneration>true</QsharpDocsGeneration>
+    <QsharpDocsPackageId>Microsoft.Quantum.Numerics</QsharpDocsPackageId>
     <IncludeQsharpCorePackages>false</IncludeQsharpCorePackages>
   </PropertyGroup>
 
   <PropertyGroup>
     <NoWarn>0162</NoWarn>
-    <RunQDocGen>True</RunQDocGen>
+    <RunQDocGen>True</RunQDocGen>https://github.com/microsoft/qsharp-runtime/pull/390
```
msoeken, line 13: That does not look like valid XML

msoeken (approved):
Thanks @cgranade. One change looked a bit odd, and I commented it. Otherwise good to go.

Numerics/src/Numerics.csproj
cgranade: It's not at all, no. I'll blame a copy-paste error. Thanks for catching it!
```suggestion
    <RunQDocGen>True</RunQDocGen>
```

PR #349, opened 2020-10-12
Make the DumpOperation operation Controllable and Adjointable

### LINKED ISSUE #336: DumpOperation is not adjointable or controllable
**Describe the bug**
`DumpOperation` (an extremely useful tool!) cannot be called from an operation that asks for all its operations to be adjointable or controllable (for example, an operation specified as `is Adj+Ctl`). This seems rather odd, since the docs say that it doesn't have an effect observable from Q#, so it would make sense to it behave like `DumpMachine` which can be called from any Q# code.

**To Reproduce**

I ran the following snippet from Q# Jupyter Notebook:
```qsharp

operation DumpCnot() : Unit is Adj+Ctl {
    DumpOperation(2, ApplyToFirstTwoQubitsCA(CNOT, _));
}
```

I got the following error:
`c:\snippet_.qs(4,5): error QS6312: The called operation does not support the necessary functor(s) for the requested auto-generation of a specialization. Missing support for functor(s) Adjoint, Controlled.`

**Expected behavior**
Operation compiled successfully and subsequent `%simulate DumpCnot` offer me a dump of CNOT.

**System information**
 - IQ# version 0.12.20092803

Pallavi-mp:
The DumpOperation should be controllable and adjointable so that 
1. It can be called from operations that are required to be controllable and adjointable.
2. It matches the behavior of other Dump* operations that are controllable and adjointable. 

Edited one existing unit test DumpCnotReturnsCorrectMatrix to call DumpCnot as a controllable and adjointable operation to validate that DumpOperation can be called from such operations. 

Resolves #336


cgranade (approved):
Thanks for adding this, @Pallavi-mp! As a minor formatting suggestion, it will likely be easier to read the signatures with a space around `+` signs. Regardless, looks like this should be good to pull in once CI runs successfully. Thanks!

Pallavi-mp:
Thanks @cgranade, I'll keep this in mind for next time. Could you also add the hacktoberfest-accepted label to this since I'm trying to participate in it? 
Also will the PR get auto-completed or would I have to check back once the validations are done to complete it?

cgranade (approved):

cgranade:
Thanks again for fixing this, @pallavi! I've marked this PR as automerge, so it should be completed once CI completes successfully. If not, I'll check back and merge it manually. Thanks!

chaos-ui (commented):
Is there anything left to be done on this?

PR #351, opened 2020-10-15
Add PrivateAssets="All" to IQ# package reference

rmshaffer:
This change improves the performance of loading the `Microsoft.Quantum.Standard.Visualization` package inside IQ# by telling NuGet to skip resolution of the IQ#-specific dependency.

This resolves https://github.com/microsoft/iqsharp/issues/338.


cgranade (approved):

cgranade:
Approved as I'm OK with this coming in, regardless of whether we _also_ apply a more general change in IQ#.

rmshaffer:
> Approved as I'm OK with this coming in, regardless of whether we _also_ apply a more general change in IQ#.

Makes sense to me, too!

PR #355, opened 2020-10-21
QDK chemistry UI

guenp:
Adding a prototype for adding some Python visualization tools to create an easier interface to the QDK chemistry library.

Right now I've created a separate Python package, because I don't quite understand how the current qsharp.chemistry package is organized and I wasn't able to make it work for my local setup (using `pip install -e .`)


cgranade:
Would you be OK with targeting this to a feature branch instead of main, especially since as per your comments we'll need to at least partially merge this with the existing `qsharp-chemistry` package?

Python/qdk-chemistry/jupyter_jsme/jsme/0368530FBC2A45A51E3B16FB1CA2E16E.cache.js:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
```
cgranade, line 1: It looks like a lot of machine-specific cache files got added to this PR?

guenp: This is the static source of the JSME release as downloaded from https://peter-ertl.com/jsme/

guenp:
@cgranade

> Would you be OK with targeting this to a feature branch instead of main

Agreed, as per our discussion in our team sync a few weeks ago. I had not yet pushed the feature branch, this is now done

Python/qdk-chemistry/jupyter_jsme/jsme/0368530FBC2A45A51E3B16FB1CA2E16E.cache.js:Zone.Identifier
cgranade: Some of these files have your username and machine name in them, though, such that I suspect there's at least something machine-specific that got caught?

guenp: Ah I see, you're right. I'll delete them. Thanks for the heads up!

guenp:
We discussed and will keep this in `qdk_chemistry` for now until we decide on a common strategy for Python packages for the QDK that are not related to Q# interoperability.

Python/qdk-chemistry/qdk_chemistry/convert.py
```diff
@@ -0,0 +1,79 @@
...
+    if charge != 0:
+        result.extend([
+            "$set",
+            f"chrg {str(charge)}",
+            "$end"
+        ])
+
+    return "\n".join(result)
+
+def get_conformer(mol: "Mol", num_confs=10) -> "Conformer":
```
msoeken, line 38: This syntax is a little bit confusing to me, but this can also be because I am not extremely familiar with Python. Is it possible to specify the precise type for `mol` and the return type? On first sight, this looks to me as if `"Mol"` is the default value for `mol`.

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
```diff
@@ -0,0 +1,136 @@
...
+
+# Rel file path for Javascript source
+JS_SOURCE = "jsme/jsme.nocache.js"
+
+# Base URL for widget source javascript
+BASE_URL = "/nbextensions/jupyter_jsme"
+
+if check_nbextension("jupyter_jsme") is False:
+    # Fallback option in case user didn't install Jupyter extension
+    BASE_URL = "https://peter-ertl.com/jsme/JSME_2020-06-11"
```
msoeken, line 24: Could we provide a fallback URL from our side?

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
```diff
@@ -0,0 +1,136 @@
...
+            height=self.height,
+            options=self.options,
+            data_type=data_type,
+            data=data
+        )
+
+    def _repr_html_(self):
+        return self.html_str
+
+    def to_mol(self, add_hs: bool=False, num_confs: int = 10) -> "Mol":
```
msoeken, line 113:
```suggestion
    def to_mol(self, add_hs: bool=False, num_confs: int=10) -> "Mol":
```

Python/qdk-chemistry/qdk_chemistry/widgets/jsmol_widget.py
```diff
@@ -0,0 +1,25 @@
+from ..convert import mol_to_xyz
+
+from jupyter_jsmol import JsmolView
+
+
+class JsmolWidget(JsmolView):
+    """Jupyter widget for JSMol molecular geometry visualization
+    """
+    def __init__(self, *args, **kwargs):
+        self.default_info['color']='white'
```
msoeken, line 10:
```suggestion
        self.default_info['color'] = 'white'
```

Python/qdk-chemistry/qdk_chemistry/widgets/jsmol_widget.py
```diff
@@ -0,0 +1,25 @@
...
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def from_mol(cls, mol: "Mol", num_confs: int = 10, *args, **kwargs):
```
msoeken, line 14:
```suggestion
    def from_mol(cls, mol: "Mol", num_confs: int=10, *args, **kwargs):
```

Python/qdk-chemistry/setup.py
```diff
@@ -0,0 +1,46 @@
+#!/bin/env python
+# -*- coding: utf-8 -*-
+##
+# setup.py: Installs Python host functionality for Q# chemistry library.
+##
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
msoeken, line 6:
```suggestion
# Copyright (c) Microsoft Corporation.
```

msoeken (approved):
I assumed that the files in `jupyter_jsme` are somehow auto generated? Other than that it looks good to me. Thanks.

msoeken:
I am wondering why there are not many CI checks? Could it be because the target branch is neither `main` nor `feature/*`?

guenp:
> I am wondering why there are not many CI checks? Could it be because the target branch is neither main nor feature/*?

Thanks for pointing that out, should be fixed now!

Python/qdk-chemistry/qdk_chemistry/convert.py
guenp: This syntax is to support third-party type checking, for example using an editor: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING 
I agree it's a bit confusing, the difference here is the colon, but adding a type hint for `num_confs` might also make this a bit more clear.

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
guenp: That's a good idea, I'm not sure where we would host it though. I think it would actually be best to create a proper notebook extension (such as https://github.com/jupyter-widgets/widget-cookiecutter).

PR #356, opened 2020-10-22
Refactor qsharp.chemistry

guenp:
Proposal for making the qsharp.chemistry library a bit easier to work with for local development (using `pip install -e .`). I renamed it to `qsharp_chemistry` and add imports to `qsharp/chemistry/__init__.py` for backward compatibility with the `qsharp.chemistry` namespace package. I also refactored the code and spit it into separate modules. Lastly, I added some notes to the README about how to install.


Python/qsharp-chemistry/setup.py
```diff
@@ -46,15 +46,15 @@
 ## SETUPTOOLS INVOCATION ##
 
 setuptools.setup(
-    name="qsharp-chemistry",
+    name="qsharp_chemistry",
```
guenp, line 49: Not sure if this change is relevant or not... It seems to install it as `qsharp-chemistry` anyways

msoeken (commented):
Thanks @guenp, looks good to me. I left some comments, but cannot make any review based on the restructuring.

Python/qsharp-chemistry/README.md
```diff
@@ -19,13 +19,29 @@ conda install -c quantum-engineering qsharp
...
+```bash
+pip install qsharp
+dotnet tool install -g Microsoft.Quantum.IQSharp
+dotnet iqsharp install
+```
+
+or simply
+
+```bash
+conda install -c quantumengineering qsharp
```
rmshaffer, line 40:
```suggestion
conda install -c quantum-engineering qsharp
```

Python/qsharp-chemistry/qsharp_chemistry/broombridge.py
```diff
@@ -0,0 +1,18 @@
+from typing import Dict
+from .problem_description import ProblemDescription
+from .fermion_hamiltonian import FermionHamiltonian
+
+class Broombridge(object):
+    """
+    Represents an instance of a broombridge schema data
```
rmshaffer, line 7:
```suggestion
    Represents an instance of a Broombridge data structure
```

Python/qsharp-chemistry/qsharp_chemistry/broombridge.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    def __init__(self, data: Dict):
+        self.__dict__ = data
+        # translate problem_descriptions into the actual datastructure:
+        self.problem_description = [ ProblemDescription(p) for p in data["problem_description"] ]
+
+    def __eq__(self, other):
+        if not isinstance(other, FermionHamiltonian):
```
rmshaffer, line 15:
```suggestion
        if not isinstance(other, Broombridge):
```

Python/qsharp-chemistry/qsharp_chemistry/fermion_hamiltonian.py
```diff
@@ -0,0 +1,45 @@
...
+        if not isinstance(other, FermionHamiltonian):
+            # don't attempt to compare against unrelated types
+            return NotImplemented
+        return self.__dict__ == other.__dict__
+
+    def add_terms(self, fermion_terms : Iterable[Tuple[List[int], float]]) -> None:
+        """
+        Adds terms to the fermion Hamiltonian.
+        """
+        import qsharp
```
rmshaffer, line 23: Is there a reason you import `qsharp` here and not at the top of the file?

Python/qsharp-chemistry/qsharp_chemistry/fermion_hamiltonian.py
```diff
@@ -0,0 +1,45 @@
...
+        result = qsharp.client._execute_magic(
+            'chemistry.fh.add_terms',
+            raise_on_stderr=True,
+            hamiltonian=self.__dict__,
+            fermion_terms=fermion_terms
+        )
+        self.__dict__ = result
+
+
+class InputState(object):
```
rmshaffer, line 34: Would it make more sense for `InputState` to live in `problem_description.py`, since that's where it's used? Probably doesn't matter too much, just a random thought.

Python/qsharp/qsharp/chemistry/__init__.py
```diff
@@ -0,0 +1,4 @@
+from qsharp_chemistry.shims import load_input_state, load_fermion_hamiltonian, load_broombridge, encode
+from qsharp_chemistry.broombridge import Broombridge
+from qsharp_chemistry.fermion_hamiltonian import FermionHamiltonian
+from qsharp_chemistry.problem_description import ProblemDescription, InputState, IndexConvention
```
rmshaffer, line 4: Right now, isn't `InputState` in `fermion_hamiltonian.py`? (Although per my other comment, perhaps `problem_description.py` is the right place anyway.)

Python/qsharp-chemistry/qsharp_chemistry/problem_description.py
```diff
@@ -0,0 +1,62 @@
...
+        if not isinstance(other, ProblemDescription):
+            # don't attempt to compare against unrelated types
+            return NotImplemented
+        return self.__dict__ == other.__dict__
+
+    def load_fermion_hamiltonian(self, index_convention : IndexConvention = IndexConvention.UpDown) -> FermionHamiltonian:
+        """
+        Loads the fermion Hamiltonian associated with this electronic structure problem.
+
+        `index_convention` can be 'UpDown' or 'HalfUp'
```
rmshaffer, line 35:
```suggestion
        `index_convention` can be `UpDown` or `HalfUp`
```

Python/qsharp/README.md
```diff
@@ -17,13 +17,32 @@ conda install -c quantum-engineering qsharp
 ## Installing from Source ##
 
 If you'd like to contribute to or experiment with the Python interoperability feature, it may be useful to install from source rather than from the `qsharp` package on the Python Package Index (PyPI).
-To do so, make sure that you are in the `Python/qsharp` directory, and run `setup.py` with the `install` argument:
+
+This package uses [namespace packages](https://www.python.org/dev/peps/pep-0382/) that are hosted in two metapackages: `qsharp-core` (part of [iqsharp](http://www.github.com/microsoft/iqsharp)) and `qsharp-chemistry` (part of [QuantumLibraries](http://www.github.com/microsoft/QuantumLibraries)). This means that to work in development mode, we need to install both.
```
rmshaffer, line 21: This is a really helpful introduction to the package structure! Super nit-pick, but since the rest of this is written in second-person:
```suggestion
This package uses [namespace packages](https://www.python.org/dev/peps/pep-0382/) that are hosted in two metapackages: `qsharp-core` (part of [iqsharp](http://www.github.com/microsoft/iqsharp)) and `qsharp-chemistry` (part of [QuantumLibraries](http://www.github.com/microsoft/QuantumLibraries)). This means that to work in development mode, you need to install both.
```

Python/qsharp-chemistry/qsharp_chemistry/shims.py
```diff
@@ -0,0 +1,88 @@
...
+import qsharp
+
+from .broombridge import Broombridge
+from .fermion_hamiltonian import FermionHamiltonian, InputState
+from .problem_description import IndexConvention
+
+import logging
+logger = logging.getLogger(__name__)
+
+_CHEMISTRY_NAMESPACE = "Microsoft.Quantum.Chemistry.Jupyter"
```
rmshaffer, line 16: Probably insignificant, but technically, in this context, this is a NuGet package name rather than a namespace. Could call this `_CHEMISTRY_PACKAGE_NAME` or something similar.

rmshaffer (commented):
I think this is a nice improvement in readability/understandability of the package source! I left just a few minor comments/suggestions, nothing too significant.

Just for clarity, this shouldn't introduce any backward-compatibility issues, right? The functionality is still identical for anyone who just does `pip install qsharp`?

Python/qsharp-chemistry/qsharp_chemistry/fermion_hamiltonian.py
guenp: Yes - I want to delay actually loading the Q# kernel until I want to use it. Maybe it's better to solve this differently.

Python/qsharp-chemistry/qsharp_chemistry/broombridge.py
guenp: Nice catch, thanks!

Python/qsharp-chemistry/qsharp_chemistry/fermion_hamiltonian.py
rmshaffer: No, that sounds perfectly reasonable to me! Just was curious.

guenp:
Thanks @msoeken and @rmshaffer for your comments!
> Just for clarity, this shouldn't introduce any backward-compatibility issues, right? The functionality is still identical for anyone who just does pip install qsharp?

Yes, I made `qsharp_chemistry` a separate package and still kept the namespace package structure in `qsharp`. So `pip install qsharp` should also install `qsharp_chemistry` and expose it through `qsharp.chemistry` (see [`qsharp/chemistry/__init__.py`](https://github.com/microsoft/QuantumLibraries/pull/356/files#diff-9858b20c75cba7801c0a12e44bfcd6c6751f6fa165eb50e84e251aee004a78b0)).

rmshaffer (approved):
This all looks good to me!

guenp:
@rmshaffer thanks for your review! I've made some changes to fix some bugs in the packaging, notably:
- create a `util.py` module that contains the functions that are currently implemented in the `iqsharp` repo, and create functions `init_qsharp` and `init_meta_path` that contain the startup script (see [`qsharp-core/qsharp/__init__.py`](https://github.com/microsoft/iqsharp/blob/86a17254c3b265603f03beaf0808c1caa01cc217/src/Python/qsharp-core/qsharp/__init__.py))
- run the startup script(/functions) in `qsharp/qsharp/__init__.py`

I've had to move this functionality over to this repo because the way I've changed the namespace packaging; by adding an `__init__.py` file here, it takes precedence over the `__init__.py` file in the `qsharp-core` repo. One way we can reduce redundancy here is to move the `util.py` file to `qsharp-core` and have identical `__init__.py` files. I suggest to merge this PR but open a PR to `iqsharp` that includes said changes, and afterwards remove the `util.py` from this feature branch.

rmshaffer:
> One way we can reduce redundancy here is to move the `util.py` file to `qsharp-core` and have identical `__init__.py` files. I suggest to merge this PR but open a PR to `iqsharp` that includes said changes, and afterwards remove the `util.py` from this feature branch.

I think that makes sense! Yes, I agree, for the feature branch this is fine temporarily but it'd be good to do that before this goes to `main`.

rmshaffer (approved):

PR #357, opened 2020-10-22
Update manifest for signed builds

anpaz:


Build/manifest.ps1
```diff
@@ -1,16 +1,31 @@
-#!/usr/bin/env pwsh
-#Requires -PSEdition Core
```
cgranade, line 2: Out of curiosity, what motivates dropping these lines?

cgranade (approved):

Build/manifest.ps1
```diff
@@ -1,16 +1,31 @@
...
-@{
+$artifacts = @{
     Packages = @(
         "Microsoft.Quantum.Standard",
         "Microsoft.Quantum.Standard.Visualization",
         "Microsoft.Quantum.Chemistry",
         "Microsoft.Quantum.Numerics",
         "Microsoft.Quantum.MachineLearning"
-    );
+    ) | ForEach-Object { Join-Path $Env:NUGET_OUTDIR "$_.$Env:NUGET_VERSION.nupkg" };
```
rmshaffer, line 27: Similar question about `.snupkg` files here.
https://github.com/microsoft/qsharp-runtime/pull/406/files#r510330957

rmshaffer, line 27: Do the `Microsoft.Quantum.Chemistry.*` subpackages also need to be listed here? I believe they are `.DataModel`, `.Jupyter`, `.Runtime`, and `.Tools`.

Build/manifest.ps1
anpaz: I don't have pwsh core installed, and plain powershell will fail if this is there.

Build/manifest.ps1
anpaz: yes, they should.

anpaz: As mentioned, they probably should but practically are not needed since we don't really use the manifest to sign packages (we just sign everything in the drop folder), so for now I'm just going to leave it like that.

Issue #335, opened 2020-09-30
Quantum addition using ApplyAnd


msoeken:
**Describe the solution you'd like**
A ripple-carry adder using the `ApplyAnd` or `ApplyLowDepthAnd` operations as described in Fig. 1 in [Halving the cost of quantum addition](https://arxiv.org/abs/1709.06648).

**Describe alternatives you've considered**
Updating `RippleCarryAdderD` to use `ApplyAnd`.

TviNet:
Hey is this issue about replacing the CCNOT gates with ApplyAnd in RippleCarryAdderD or implementing the circuit given in Fig. 1 which does not compute the Carry out bit?

msoeken:
I suppose that `RippleCarryAdderD` could be updated. However, it's not just replacing existing `CCNOT` since the current implementation used two `CCNOT` for each `Carry`, where only one is necessary. It also should support a `Carry` bit, unlike the figure in the paper.

PR #358, opened 2020-10-25
RippleCarryAdderD update

TviNet:
PR for issue #335
Implemented controlled adder blocks following [this paper](https://arxiv.org/abs/1709.06648 ) but use the sum and carry blocks to compute the output carry bit.

### DIFF
--- a/Standard/src/Arithmetic/Integer.qs
+++ b/Standard/src/Arithmetic/Integer.qs
@@ -31,8 +31,28 @@ namespace Microsoft.Quantum.Arithmetic {
     }
 
     /// # Summary
-    /// Implements an adder block that uses a temporary logical-AND construction.
-    /// The carryOut is computed and then erased using the adjoint operation. 
+    /// Implements a compute carry gate using a temporary logical-AND construction.
+    ///
+    /// # Input
+    /// ## carryIn
+    /// Carry-in qubit.
+    /// ## summand1
+    /// First summand qubit.
+    /// ## summand2
+    /// Second summand qubit.
+    /// `summand1` and `summand2`.
+    /// ## carryOut
+    /// Carry-out qubit, will be xored with the higher bit of the sum.
+    internal operation ComputeTemporaryCarry(carryIn: Qubit, summand1: Qubit, summand2: Qubit, carryOut: Qubit)
+    : Unit is Adj{
+        CNOT(carryIn, summand2);
+        CNOT(carryIn, summand1);
+        ApplyAnd(summand1, summand2, carryOut);
+        CNOT(carryIn, carryOut);
+    }
+
+    /// # Summary
+    /// Implements a erase carry gate by uncomputing the logical-AND.
     ///
     /// # Input
     /// ## carryIn
@@ -43,22 +63,17 @@ namespace Microsoft.Quantum.Arithmetic {
     /// Second summand qubit, is replaced with the lower bit of the sum of
     /// `summand1` and `summand2`.
     /// ## carryOut
-    /// Carry-out qubit, will be xored with the higher bit of the sum.
-    operation AdderBlockUsingTemporaryLogicalAnd(carryIn: Qubit, summand1: Qubit, summand2: Qubit, carryOut: Qubit)
-    : Unit is Adj + Ctl {
-        body (...) {
-            CNOT(carryIn, summand2);
+    /// Carry-out qubit, will be erased
+    internal operation EraseTemporaryCarry(carryIn: Qubit, summand1: Qubit, summand2: Qubit, carryOut: Qubit)
+    : Unit is Adj + Ctl{
+        body (...){
+            CNOT (carryIn, carryOut);
+            (Adjoint ApplyAnd) (summand1, summand2, carryOut);
+            CNOT (summand1, summand2);
             CNOT(carryIn, summand1);
-            ApplyAnd(summand1, summand2, carryOut);
-            CNOT(carryIn, carryOut);
-        }
+            CNOT(carryIn, summand2);
+		}
         controlled (controls, ...) {
-            CNOT(carryIn, summand2);
-            CNOT(carryIn, summand1);
-            ApplyAnd(summand1, summand2, carryOut);
-            CNOT(carryIn, carryOut);
-        }
-        controlled adjoint (controls, ...) {
             CNOT (carryIn, carryOut);
             (Adjoint ApplyAnd) (summand1, summand2, carryOut);
             (Controlled CNOT) (controls, (summand1, summand2));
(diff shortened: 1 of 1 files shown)


Standard/src/Arithmetic/Integer.qs
```diff
@@ -30,6 +30,43 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// ## carryIn
+    /// Carry-in qubit.
+    /// ## summand1
+    /// First summand qubit.
+    /// ## summand2
+    /// Second summand qubit, is replaced with the lower bit of the sum of
+    /// `summand1` and `summand2`.
+    /// ## carryOut
+    /// Carry-out qubit, will be xored with the higher bit of the sum.
+    operation AdderBlockUsingTemporaryLogicalAnd(carryIn: Qubit, summand1: Qubit, summand2: Qubit, carryOut: Qubit)
```
msoeken, line 47: This operation can be declared `internal` as it is only used for the implementation of `RippleCarryAdderD`.

Standard/src/Arithmetic/Integer.qs
```diff
@@ -30,6 +30,43 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    : Unit is Adj + Ctl {
```
msoeken, line 48: This operation does not need to be `Ctl`, since it is only called inside a `within` block. Then the controlled blocks can be removed and also the `body (...)` line can be removed, as it will be the only specialisation (the `adjoint` can be generated automatically).

Standard/src/Arithmetic/Integer.qs
```diff
@@ -90,17 +129,23 @@ namespace Microsoft.Quantum.Arithmetic {
                 "Input registers must have the same number of qubits."
             );
 
-            using (auxRegister = Qubit[nQubits]) {
-                for (idx in 0..(nQubits-2)) {
-                    Carry(auxRegister[idx], xs![idx], ys![idx], auxRegister[idx+1]);           // (1)
+            using (auxRegister = Qubit[nQubits-1]) {
+                within { 
+                    ApplyAnd (xs![0], ys![0], auxRegister[0]);
```
msoeken, line 134: It should be possible to merge these nested `within / apply` blocks by moving this `ApplyAnd` operation call before the `for` loop in the nested `within` block.

Standard/src/Arithmetic/Integer.qs
```diff
@@ -90,17 +129,23 @@ namespace Microsoft.Quantum.Arithmetic {
...
-                (Controlled Carry) (controls, (auxRegister[nQubits-1], xs![nQubits-1], ys![nQubits-1], carry));
-                (Controlled CNOT) (controls, (xs![nQubits-1], ys![nQubits-1]));
-                (Controlled Sum) (controls, (auxRegister[nQubits-1], xs![nQubits-1], ys![nQubits-1]));
-                for (idx in (nQubits-2)..(-1)..0 ) {
-                    (Adjoint Carry) (auxRegister[idx], xs![idx], ys![idx], auxRegister[idx+1]); // cancels with (1)
-                    (Controlled Sum) (controls, (auxRegister[idx], xs![idx], ys![idx]));
+                apply {
+                    within {
+                        for (idx in 1..(nQubits-2)) {
+                            (Controlled AdderBlockUsingTemporaryLogicalAnd) (controls, (auxRegister[idx-1], xs![idx], ys![idx], auxRegister[idx]));
```
msoeken, line 139:
```suggestion
                            AdderBlockUsingTemporaryLogicalAnd(auxRegister[idx-1], xs![idx], ys![idx], auxRegister[idx]);
```

This statement does not need to be controlled, as it is inside a `within` block.

Standard/src/Arithmetic/Integer.qs
```diff
@@ -90,17 +129,23 @@ namespace Microsoft.Quantum.Arithmetic {
...
+                        }
+                    }
+                    apply {
+                        (Controlled Carry) (controls, (auxRegister[nQubits-2], xs![nQubits-1], ys![nQubits-1], carry));
```
msoeken, line 143: This `Carry` implementation uses two `CCNOT` gates and could be replaced by something that uses a single `ApplyAnd`. Could `AdderBlockUsingTemporaryLogicalAnd` be used?

Standard/src/Arithmetic/Integer.qs
```diff
@@ -90,17 +129,23 @@ namespace Microsoft.Quantum.Arithmetic {
...
+                        (Controlled CNOT) (controls, (xs![nQubits-1], ys![nQubits-1]));
+                        (Controlled Sum) (controls, (auxRegister[nQubits-2], xs![nQubits-1], ys![nQubits-1]));
+                    }
                 }
+                (Controlled CNOT) (controls,(xs![0], ys![0]));
```
msoeken, line 148: It is not clear to me why this line is here. Could you add some line of documentation to make it clearer.

msoeken (changes requested):
Thanks a lot for your pull request @TviNet. It looks already very good. I left some comments.

Standard/src/Arithmetic/Integer.qs
TviNet: Thanks for the suggestions
I am using this figure as reference.
image

The block can be considered to have two parts: ComputeTempCarry and EraseTempCarry.
 To make use of within..apply structure I have used
 ```
AdderBlockUsingTemporaryLogicalAnd =  ComputeTempCarry 
Adjoint AdderBlockUsingTemporaryLogicalAnd  = EraseTempCarry
```
Since the control only affects EraseTempCarry I had to add a Controlled + Adjoint variant.

I realize this has gotten a bit messy and is probably an incorrect usage of adjoint 

So I am planning to use the two blocks (ComputeTempCarry is Adj) and (EraseTempCarry is Adj + Ctl) but this would not make use of within..apply

File: Standard/src/Arithmetic/Integer.qs (hunk 5 above)
Comment (TviNet) [reply Standard/src/Arithmetic/Integer.qs:143]: `AdderBlockUsingTemporaryLogicalAnd` has to erase the carry bit by design so I don't think it can be used here unless I understand incorrectly.  The first `CCNOT` of  `Carry` can be changed to `ApplyAnd` since the `CarryOut` is zero. I haven't been able to find another implementation that would do better.

File: Standard/src/Arithmetic/Integer.qs (hunk 6 above)
Comment (TviNet) [reply Standard/src/Arithmetic/Integer.qs:148]: Since the first bit has `CarryIn` always zero the operations on first bit can be simplified to 
```
ApplyAnd (xs![0], ys![0], auxRegister[0]);
---<Compute other bits using auxRegister[0]>---
(Adjoint ApplyAnd) (xs![0], ys![0], auxRegister[0]);
(Controlled CNOT) (controls,(xs![0], ys![0]));
```
to reduce the gate count.

msoeken (approved):
Thanks for your changes and the clarification. I approve this PR and changed the target to a feature branch.

Thanks for your work on this @TviNet

PR #360, opened 2020-10-28
Chemistry data wrangling tools

guenp:
Add UTs and functionality to create molecular geometry and export to/import from XYZ format and to NWChem format.

To do in follow-up PR:

- [ ] Add support for OpenMolcas and Psi4
- [ ] Use `Geometry` consistently (instead of `coordinates` tuples)

Done:
- [x] "Save" button in JSME widget does conversions under the hood

Other:
[-] Issue with Jupyter hotkeys interfering with JS applet (e.g. "a", "b", "d" etc.)
Won't fix - This would require disabling/enabling the [shortcuts ](https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Custom%20Keyboard%20Shortcuts.html)which I think is not a great idea; instead, one can now just pass SMILES to the widget directly.


Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,116 @@
...
+    Args:
+        xyz (str): XYZ file format
+    """
+    lines = xyz.split("\n")
+    return "\n".join(lines[2:])
+
+def create_input_deck(
+        mol_name: str, 
+        geometry: str, 
+        num_orbitals: int,
```
dgwvfxn, line 65: I suggest to use 'num_active_orb' to avoid confusion.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,116 @@
...
+        memory: str = "memory stack 1000 mb heap 100 mb global 1000 mb noverify",
+        geometry_units: str = "au", 
+        basis: str = "sto-3g",
+        charge: int = 0,
+        rhf: str = "rhf",
+        spin: str = "singlet",
+        method: str = "ccsd",
+        driver: str = "energy",
+        mol: "Mol" = None,
+        num_el: int = None,
```
dgwvfxn, line 75: I suggest to use 'num_active_el' to avoid confusion.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,116 @@
...
+nroots 5
+thresh 1.0e-6
+end
+
+set tce:print_integrals T
+set tce:qorb {num_orb}
+set tce:qela {num_el_a}
+set tce:qelb {num_el_b}
+
+task tce energy
```
dgwvfxn, line 50: Should this be `task tce {driver}`? I didn't see the `driver` keyword in the template

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,116 @@
...
+        mol_name (str): Molecule name
+        geometry (str): Molecule geometry in the following format (each atom on a new line):
+            [Atom] [X] [Y] [Z]
+        num_orbitals (int): Number of orbitals in molecular ground state to set active space
+        memory (str, optional): Memory specification. Defaults to "memory stack 1000 mb heap 100 mb global 1000 mb noverify".
+        geometry_units (str, optional): Units used for geometry. Defaults to "au".
+        basis (str, optional): Basis to use for atoms. Defaults to "sto-3g".
+        charge (int, optional): Molecule charge. Defaults to 0.
+        rhf (str, optional): Restricted Hartree Fock method. Either "rhf" or "rohf". Defaults to "rhf".
+        spin (str, optional): Spin property. Defaults to "singlet".
```
dgwvfxn, line 89: For `spin` variable, add the following explanation: `nopen N` is also a supported format, N is the number of singly occupied electrons.

dgwvfxn (commented):
Looks great! I have some comments to improve the naming of varaibles. Looks like {driver} key is missing from the nwchem template.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,116 @@
...
+{rhf}
+{spin}
+end
+
+tce
+{method}
+tilesize 1
+2eorb
+2emet 13
+nroots 5
```
dgwvfxn, line 41: We should also expose the {nroot} to user as num_tce_root, and {thresh} to user as tce_threshold.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: That should be there, thanks for catching that

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: I'll add in a new variable for that.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,144 @@
...
+        num_active_orbitals: int,
+        memory: str = "memory stack 1000 mb heap 100 mb global 1000 mb noverify",
+        geometry_units: str = "au", 
+        basis: str = "sto-3g",
+        charge: int = 0,
+        scf_thresh: float = 1.0e-10,
+        scf_tol2e: float = 1.0e-10,
+        rhf: str = "rhf",
+        spin: str = "singlet",
+        nopen: int = None,
```
cgranade, line 79: Elsewhere we've used `n_` as a prefix for "number of," and the rest of the PR uses `num_`. Does this use a different convention to match NWChem?

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Args:
+        xyz (str): XYZ file format
+    """
+    match = re.findall(XYZ_PATTERN, xyz)
+    return "\n".join(" ".join(item) for item in match)
+
+def create_input_deck(
+        mol_name: str, 
+        geometry: str,
```
cgranade, line 69: It looks like the user should pass in the NWChem-formatted geometry block, e.g. as generated by `geometry_from_xyz` above. Would it make sense to allow the user to pass geometry in a more Python-ic way as well, such as a `ndarray` of dtype `float` and shape `(n, 3)`? Could even use `Union[str, np.ndarray]` type hint to allow users to still provide NWChem syntax directly.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,144 @@
...
+        num_active_el (int, optional): Number of active electrons in molecule. This value is 
+            calculated based on atomic numbers if mol is provided. Defaults to None.
+
+    Returns:
+        str: NW Chem input deck formatted string
+    """
+    if mol is not None and num_active_el is None:
+        num_active_el = num_electrons(mol)
+    elif mol is None and num_active_el is None:
+        raise ValueError("Cannot proceed: please provide either a Mol object or specify number of electrons")
```
cgranade, line 119: What happens if `mol` and `num_active_el` are both not `None`? It looks like `mol` is ignored in that case; would it be worth documenting that explicitly (I think it's implied at L108), or perhaps raising a warning in that case?

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,144 @@
...
+        xyz (str): XYZ file format
+    """
+    match = re.findall(XYZ_PATTERN, xyz)
+    return "\n".join(" ".join(item) for item in match)
+
+def create_input_deck(
+        mol_name: str, 
+        geometry: str, 
+        num_active_orbitals: int,
+        memory: str = "memory stack 1000 mb heap 100 mb global 1000 mb noverify",
```
cgranade, line 71: It looks like values for this argument expose a lot of NWChem's syntax for memory management; can we provide a link to help with that, or perhaps abstract with a `@dataclass` that represents the information needed for NWChem's `memory` line?

cgranade (commented):
I recognize this is in draft, so this is just some early feedback in case that's useful; please let me know when you'd like for me to take a more detailed look. Thank you!

Python/qdk-chemistry/qdk_chemistry/convert.py
```diff
@@ -79,3 +80,15 @@ def mol_to_xyz(mol: "Mol", num_confs: int=10) -> str:
     conformer = get_conformer(mol=mol, num_confs=num_confs)
 
     return conformer_to_xyz(number_of_atoms, charge, symbols, conformer)
+
+
+def num_electrons(mol: "Mol") -> int:
+    """Calculate the number of electrons in the molecule
+
+    Args:
```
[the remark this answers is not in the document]
guenp: Ah, I suspected this was the case, for now I went with the default settings of the VSCode Python Docstring Generator but this is great to know :) thanks for the link!

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,144 @@
...
+set tce:qelb {num_el_b}
+
+task tce {driver}
+"""
+
+FLOAT_PATTERN = "([+-]?[0-9]*[.][0-9]+)"
+XYZ_PATTERN = f"(\w) {FLOAT_PATTERN} {FLOAT_PATTERN} {FLOAT_PATTERN}"
+
+def geometry_from_xyz(xyz: str):
+    """Generate geometry portion of NW Chem file from XYZ data
```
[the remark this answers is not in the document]
guenp: Sure thing! I'll add in a link to the Wikipedia article. Adding an example is also a good idea.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    match = re.findall(XYZ_PATTERN, xyz)
+    return "\n".join(" ".join(item) for item in match)
+
+def create_input_deck(
+        mol_name: str, 
+        geometry: str, 
+        num_active_orbitals: int,
+        memory: str = "memory stack 1000 mb heap 100 mb global 1000 mb noverify",
+        geometry_units: str = "au",
```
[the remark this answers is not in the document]
guenp: Absolutely, this is one of the reasons why this is still a draft! :) I don't (yet) intend to make a full-fledged Python builder for NWChem, just one that is able to deal with some of the samples we want to work with. But this is a great suggestion, thanks.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: Yes, this is meant to echo the NWChem syntax.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: Yes, I haven't decided yet what format to use; I used a numpy array before and found that a bit untransparent. I liked your suggestion earlier of using dataclasses. Perhaps just a list of namedtuples will do the trick.

cgranade: If you want the power of NumPy arrays but with something closer to dataclasses or namedtuples, another option could be record arrays. E.g.: `dtype=[('x', float), ('y', float), ('z', float)]`. That said, I would guess that most geometries are small enough compared to all the other data going about that a list of namedtuples or dataclasses would indeed work well, and would play better with type hinting too (I don't think you can constrain ndarray typehints by dtype?).

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: As mentioned earlier, I don't really want to build a comprehensive nwchem builder here. This is really just meant as a handy tool to generate NWChem on-the-fly. For advanced usage, the user still needs to know/understand NWChem syntax.

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: I added in a warning, thanks for catching this!

Python/qdk-chemistry/qdk_chemistry/solvers/nwchem.py
guenp: I went with a combination of a `dataclass` and a subclass of `list`, because I want the coordinates to be mutable if necessary.

guenp:
@cgranade  this is ready for review! I will follow up with another PR with support for other quantum chemistry libraries.

Python/qdk-chemistry/qdk_chemistry/tests/conftest.py
```diff
@@ -0,0 +1,21 @@
+import pytest
```
cgranade, line 1: I suspect the test modules will also need ⓒ headers.

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
```diff
@@ -101,38 +135,57 @@ def html_str(self) -> str:
...
-            Mol: RDKit Mol object
+        :param add_hs: Add Hydrogen atoms
+        :type add_hs: bool
+        :param num_confs: Number of conformers to generate
+        :type num_confs: int
+        :return: RDKit molecule object
+        :rtype: Mol
         """
-        if self.value.smiles is not "":
+        if self.value.smiles != "":
```
cgranade, line 174: Since the empty string is the only false-y string, you may be able to replace:

```suggestion
        if self.value.smiles:
```

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
```diff
@@ -101,38 +135,57 @@ def html_str(self) -> str:
...
             mol = Chem.MolFromSmiles(self.value.smiles)
-        elif self.value.molblock is not "":
+        elif self.value.molblock != "":
             mol = Chem.MolFromMolBlock(self.value.molblock)
         else:
             raise ValueError("Cannot create Mol object: JSME value is empty")
 
-        if add_hs is True:
+        if add_hs is True and mol is not None:
```
cgranade, line 181:
```suggestion
        if add_hs and mol is not None:
```

cgranade (commented):
Looking good, thanks for doing this. Had a few comments, mostly about defining the exact API surface for this tool and what parts a user will be expected to work with directly.

Python/qdk-chemistry/qdk_chemistry/geometry/geometry.py
```diff
@@ -0,0 +1,107 @@
...
+
+        :return: List of coordinate tuples
+        :rtype: Tuple[str, float, float, float]
+        """
+        for element in self:
+            yield ( element.name, element.x, element.y, element.z )
+
+    @classmethod
+    def from_mol(cls, mol: "Mol", num_confs: int = 10):
+        """Constructor for creating geometry from RDKit molecule
```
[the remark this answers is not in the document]
guenp: I don't think it's reserved in docstrings? In it's pure definition a constructor is a method that returns an instance of the class. I can rephrase this a bit to avoid confusion.

Python/qdk-chemistry/qdk_chemistry/geometry/geometry.py
```diff
@@ -0,0 +1,107 @@
...
+            <element> <X> <Y> <Z>
+            ...
+
+        Source: https://en.wikipedia.org/wiki/XYZ_file_format.
+
+        :param xyz: XYZ file format
+        :type xyz: str
+        """
+        match = re.findall(XYZ_PATTERN, xyz)
+        return cls([Element.from_tuple(item) for item in match])
```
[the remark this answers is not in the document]
guenp: I don't think that's a failure mode we should support. If the XYZ file is properly formatted it should not have XYZ data in the header. I do think it will be a good idea to add an assertion that the number of atoms is correct.

Python/qdk-chemistry/qdk_chemistry/geometry/xyz.py
```diff
@@ -0,0 +1,48 @@
+from typing import Iterable, Tuple
+
+def element_to_xyz(name: str, x: float, y: float, z: float) -> str:
```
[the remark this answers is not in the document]
guenp: Yes these are intended for public use.

Python/qdk-chemistry/qdk_chemistry/geometry/rdkit_convert.py
```diff
@@ -0,0 +1,127 @@
...
+        ((not_converged, energy), conformer) = min(zip(res, mol.GetConformers()), key=lambda x: x[0][1])
+        if not_converged:
+            print(f"Solution did not converge. Lowest energy found: {energy}")
+    return conformer
+
+
+def conformer_to_coordinates(
+        symbols: Iterable[str], 
+        conformer: "Conformer"
+    ) -> List[Tuple[str, float, float, float]]:
```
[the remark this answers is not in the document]
guenp: I guess I could refactor this into `conformer_to_geometry` and then just use `Geometry.to_xyz()`, although that seems a bit silly because it does the extra work of creating a `Geometry` object and the user wouldn't access that anyway. But as to your previous point I can make them private and make a `Geometry` version that is public.

Python/qdk-chemistry/qdk_chemistry/geometry/rdkit_convert.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+
+    return coordinates_to_xyz(
+        number_of_atoms=number_of_atoms,
+        charge=charge,
+        coordinates=coordinates
+    )
+
+
+def mol_to_coordinates(mol: "Mol", num_confs: int=10) -> List[Tuple[str, float, float, float]]:
```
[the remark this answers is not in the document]
guenp: See comments above, we can make this private as there is already the equivalent `Geometry.from_mol` method.

Python/qdk-chemistry/qdk_chemistry/tests/conftest.py
guenp: Merci

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
guenp: I strongly favor explicit over implicit; for example, if someone would set `smiles` to `False` or `0`, the above would not pass. Being explicit consistently can avoid these kinds of edge cases.

Python/qdk-chemistry/qdk_chemistry/widgets/jsme_widget.py
guenp: See above comment on explicit vs. implicit; in this case, if someone passes the wrong type accidentally, this will still pass. However, if this is set explicitly to `is True`, it will assume the route of when `add_hs` was set to `False`. I find that this is a way to avoid some gnarly situations (it will execute this code if and only if the flag is set to `True`)

Python/qdk-chemistry/qdk_chemistry/geometry/geometry.py
```diff
@@ -0,0 +1,107 @@
...
+    x: float
+    y: float
+    z: float
+
+    @classmethod
+    def from_tuple(cls, value: Tuple[str, float, float, float]):
+        name, x, y, z = value
+        return cls(name=name, x=float(x), y=float(y), z=float(z))
+
+    def to_xyz(self):
```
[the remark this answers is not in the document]
guenp: Yes, it's public; it could be handy for e.g. saving this data to an .XYZ file.

Python/qdk-chemistry/qdk_chemistry/geometry/rdkit_convert.py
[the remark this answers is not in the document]
guenp: Actually, `conformer_to_xyz` is not used, so I will just remove it

cgranade (approved):
As discussed offline, I think this should be ready for bringing into the feature branch. There may still be some refactoring to do in future PRs, but I think you've captured that in `- [ ]` items. Thanks!

PR #365, opened 2020-11-05
Consolidate Chemistry data wrangling tools into one function

guenp:
One function to load a Broombridge file and generate a tuple to pass as `JWEncodedData`.

### DIFF
--- a/Python/qsharp-chemistry/qsharp_chemistry/problem_description.py
+++ b/Python/qsharp-chemistry/qsharp_chemistry/problem_description.py
@@ -9,6 +9,7 @@ import json
 import typing
 
 from typing import List, Tuple, Dict, Iterable
+from dataclasses import dataclass, field
 from enum import Enum
 
 from .fermion_hamiltonian import FermionHamiltonian
@@ -35,10 +36,24 @@ class InputState(object):
         return self.__dict__ == other.__dict__
 
 
+@dataclass
 class ProblemDescription(object):
     """
     Represents an electronic structure problem.
     """
+    metadata: dict
+    initial_state_suggestions: list
+    basis_set: dict = field(default_factory=dict)
+    geometry: dict = field(default_factory=dict)
+    coulomb_repulsion: dict = field(default_factory=dict)
+    scf_energy: dict = field(default_factory=dict)
+    scf_energy_offset: dict = field(default_factory=dict)
+    fci_energy: dict = field(default_factory=dict)
+    n_orbitals: int = 0
+    n_electrons: int = 0
+    energy_offset: dict = field(default_factory=dict)
+    hamiltonian: dict = field(default_factory=dict)
+
     def __init__ (self, data: Dict):
         self.__dict__ = data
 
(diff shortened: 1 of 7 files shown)


Python/qsharp-chemistry/qsharp_chemistry/shims.py
```diff
@@ -93,3 +93,41 @@ def encode(hamiltonian: FermionHamiltonian, input_state: InputState) -> Tuple:
         input_state=input_state.__dict__
     )
     return data
+
+
+def load_and_encode(
+    file_name: str,
+    problem_description_index: int = 0,
+    initial_state_label: str = None
+) -> Tuple[
```
msoeken, line 102: Is it possible to define type aliases in Python? This is a little hard to parse.

msoeken (approved):
Looks good to me, just had one small comment.

Python/qsharp-chemistry/qsharp_chemistry/shims.py
```diff
@@ -93,3 +93,41 @@ def encode(hamiltonian: FermionHamiltonian, input_state: InputState) -> Tuple:
...
+    # TODO: This function accesses a file multiple times, find a way to cache 
+    # or pass data explicitly
+    # TODO: Make these tuples of lists etc. a bit more insightful data types
+    broombridge_data =  load_broombridge(file_name)
+    problem = broombridge_data.problem_description[problem_description_index]
+
+    if initial_state_label is None:
+        # Pick first in list
+        initial_state_label = problem.initial_state_suggestions[0].get("Label")
+        logger.info(f"Using inital state label: {initial_state_label}")
```
rmshaffer, line 127:
```suggestion
        logger.info(f"Using initial state label: {initial_state_label}")
```

Python/qsharp-chemistry/qsharp_chemistry/shims.py
rmshaffer: I just recently discovered: https://docs.python.org/3/library/typing.html#type-aliases

Python/qsharp-chemistry/qsharp_chemistry/problem_description.py
```diff
@@ -35,10 +36,24 @@ def __eq__(self, other):
...
+    geometry: dict = field(default_factory=dict)
+    coulomb_repulsion: dict = field(default_factory=dict)
+    scf_energy: dict = field(default_factory=dict)
+    scf_energy_offset: dict = field(default_factory=dict)
+    fci_energy: dict = field(default_factory=dict)
+    n_orbitals: int = 0
+    n_electrons: int = 0
+    energy_offset: dict = field(default_factory=dict)
+    hamiltonian: dict = field(default_factory=dict)
+
```
rmshaffer, line 56: This is a nice idea! Are these used anywhere?

rmshaffer (approved):

Python/qsharp-chemistry/qsharp_chemistry/problem_description.py
guenp: Good point, right now I just override the `__dict__` property in the constructor, which is kind of a hack... I will make sure to explicitly set each value by grabbing them from the dict.

Python/qsharp-chemistry/qsharp_chemistry/shims.py
guenp: Will do, thanks both!

PR #370, opened 2020-11-09
Enable PrepareArbitraryState to use state injection on Quantum Simulator

irinayat-MS:
Added new non-Adj/non-Ctr operation `_PrepareAmplitudesFromZeroState` that will emulate state preparation if running on a simulator that supports it and the prerequisites are met.

Had to move onto SDK 0.14.20111301-pull to pick up the new functionality in the Quantum Simulator.

### DIFF
--- a/Standard/src/Preparation/Arbitrary.cs
+++ b/Standard/src/Preparation/Arbitrary.cs
@@ -12,17 +12,17 @@ using Microsoft.Quantum.Simulation.Simulators;
 
 namespace Microsoft.Quantum.Preparation
 {
-    public partial class ApproximatelyPrepareArbitraryState
+    public partial class _PrepareAmplitudesFromZeroState
     {
         /// <summary>
         ///  Provides a native emulation of the ApproximatelyPrepareArbitraryState operation when
         ///  the operation is executed using the full-state QuantumSimulator.
         /// </summary>
-        public class Native : ApproximatelyPrepareArbitraryState
+        public class Native : _PrepareAmplitudesFromZeroState
         {
             [DllImport(QuantumSimulator.QSIM_DLL_NAME, ExactSpelling = true,
                 CallingConvention = CallingConvention.Cdecl, EntryPoint = "InjectState")]
-            private static extern double InjectState(uint sid, uint n, uint[] q, double[] re, double[] im);
+            private static extern bool InjectState(uint sid, uint n, uint[] q, double[] re, double[] im);
 
             private QuantumSimulator? Simulator { get; }
 
@@ -35,9 +35,9 @@ namespace Microsoft.Quantum.Preparation
             /// Overrides the body to do the emulation when possible. If emulation is not possible, then
             /// it just invokes the default Q# implementation.
             /// </summary>
-            public override Func<(double, IQArray<Math.ComplexPolar>, Arithmetic.LittleEndian), QVoid>__Body__ => (_args) =>
+            public override Func<(IQArray<Math.ComplexPolar>, Arithmetic.LittleEndian), QVoid>__Body__ => (_args) =>
             {
-                var (tolerance, polarAmplitudes, qubits) = _args;
+                var (polarAmplitudes, qubits) = _args;
 
                 // TODO: benchmark for small `qubits` arrays to find out in which cases emulation is actually
                 // benefitial.
@@ -46,28 +46,32 @@ namespace Microsoft.Quantum.Preparation
                     return base.__Body__(_args);
                 }
 
-                // calculate the norm as we might need to normalize the state
+                // Calculate the norm as we might need to normalize the requsted state.
                 var norm = 0.0;
                 foreach (var pa in polarAmplitudes) { norm += pa.Magnitude * pa.Magnitude; }
                 norm = System.Math.Sqrt(norm);
 
-                var state_size = polarAmplitudes.Length;
-                var re = new double[state_size];
-                var im = new double[state_size];
-                for (int i = 0; i < state_size; i++)
+                // Setup the amplitudes arrays for the call to native (it needs to translate from polar to cartesian and
+                // might need to pad the tail of an incomplete amplitudes' array with zeros).
+                var stateSize = (long)1 << (int)qubits.Data.Length;
+                var re = new double[stateSize];
+                var im = new double[stateSize];
+                for (long i = 0; i < polarAmplitudes.Length; i++)
                 {
                     var pa = polarAmplitudes[i];
                     re[i] = (System.Math.Abs(pa.Magnitude) * System.Math.Cos(pa.Argument))/norm;
                     im[i] = (System.Math.Abs(pa.Magnitude) * System.Math.Sin(pa.Argument))/norm;
                 }
-
-                // Replace implementation in the library to stop throwing and return success/failure result instead.
-                try
+                for (long i = polarAmplitudes.Length; i < stateSize; i++)
                 {
-                    // State injection is precise so tolerance isn't used.
-                    InjectState(Simulator.Id, (uint)qubits.Data.Length, qubits.Data.GetIds(), re, im);
+                    re[i] = 0.0;
+                    im[i] = 0.0;
                 }
-                catch
+
+                // Emulation might fail if the target qubits are entangled or not all in state |0>. In this case
+                // we should fallback to the quantum state preparation as it guarantees the operation to be a proper
+                // unitary no matter the state of the qubits.
+                if (!InjectState(Simulator.Id, (uint)qubits.Data.Length, qubits.Data.GetIds(), re, im))
                 {
                     return base.__Body__(_args);
                 }
--- a/Standard/src/Preparation/Arbitrary.cs
+++ b/Standard/src/Preparation/Arbitrary.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Quantum.Preparation
                 var (polarAmplitudes, qubits) = _args;
 
                 // TODO: benchmark for small `qubits` arrays to find out in which cases emulation is actually
-                // benefitial.
+                // beneficial.
                 if (this.Simulator == null)
                 {
                     return base.__Body__(_args);
--- a/Standard/src/Preparation/Arbitrary.cs
+++ b/Standard/src/Preparation/Arbitrary.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections;
+using System.Runtime.InteropServices;
+using Microsoft.Quantum.Simulation;
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators;
+
+
+namespace Microsoft.Quantum.Preparation
+{
+    public partial class PrepareArbitraryState
+    {
+        /// <summary>
+        ///  Provides a native emulation of the PrepareArbitraryState operation when
+        ///  the operation is executed using the full-state QuantumSimulator.
+        /// </summary>
+        public class Native : PrepareArbitraryState
+        {
+            [DllImport(QuantumSimulator.QSIM_DLL_NAME, ExactSpelling = true,
+                CallingConvention = CallingConvention.Cdecl, EntryPoint = "InjectState")]
+            private static extern double InjectState(uint sid, uint n, uint[] q, double[] re, double[] im);
+
+            private QuantumSimulator Simulator { get; }
+
+            public Native(IOperationFactory m) : base(m)
+            {
+                this.Simulator = m as QuantumSimulator;
+            }
+
+            /// <summary>
+            /// Overrides the body to do the emulation when possible. If emulation is not possible, then
+            /// it just invokes the default Q# implementation.
+            /// </summary>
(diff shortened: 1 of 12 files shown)


Standard/src/Preparation/Arbitrary.cs
```diff
@@ -0,0 +1,66 @@
...
+            public override Func<(IQArray<Math.ComplexPolar>, Arithmetic.LittleEndian), QVoid>__Body__ => (_args) =>
+            {
+                var (polar_amplitudes, qubits) = _args;
+
+                if (this.Simulator == null)
+                {
+                    return base.__Body__(_args);
+                }
+
+                // calculate the norm as we might need to normalize the state
```
irinayat-MS, line 46: PrepareArbitraryState allows the user provided amplitudes that aren't normalized. It means we haven't to do the math even if the user did normalize the state themselves...

Standard/src/Preparation/Arbitrary.cs
```diff
@@ -0,0 +1,66 @@
...
+            public Native(IOperationFactory m) : base(m)
+            {
+                this.Simulator = m as QuantumSimulator;
+            }
+
+            /// <summary>
+            /// Overrides the body to do the emulation when possible. If emulation is not possible, then
+            /// it just invokes the default Q# implementation.
+            /// </summary>
+            public override Func<(IQArray<Math.ComplexPolar>, Arithmetic.LittleEndian), QVoid>__Body__ => (_args) =>
```
cgranade, line 37: What's the 1-norm difference between the injected and prepared states? If possible, I'd suggest adding a unit test to check that the 1-norm is small for a few representative examples, so as to make sure that the emulation feature cannot be detected by user code other than by the relative wall-clock performance.

Standard/src/Preparation/Arbitrary.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Microsoft.Quantum.Preparation
+{
+    public partial class PrepareArbitraryState
+    {
+        /// <summary>
+        ///  Provides a native emulation of the PrepareArbitraryState operation when
+        ///  the operation is executed using the full-state QuantumSimulator.
+        /// </summary>
+        public class Native : PrepareArbitraryState
```
cgranade, line 20: One thing to watch out for is that the QML library currently doesn't call `PrepareAribtraryState`, but rather calls `_CompileApproximateArbitraryStatePreparation` directly so as to cache classical parts of state preparation calculations across repeated preparations of the same state. As a result, the operation being emulated in this PR isn't actually on the hot path for existing QML functionality.

Standard/src/Preparation/Arbitrary.cs
cgranade: It may be possible to share this logic with the Q# operation by refactoring `_CompileApproximateArbitraryStatePreparation` slightly and emulating something in its call graph instead of `PrepareArbitraryState` directly.

Standard/src/Preparation/Arbitrary.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            [DllImport(QuantumSimulator.QSIM_DLL_NAME, ExactSpelling = true,
+                CallingConvention = CallingConvention.Cdecl, EntryPoint = "InjectState")]
+            private static extern double InjectState(uint sid, uint n, uint[] q, double[] re, double[] im);
+
+            private QuantumSimulator Simulator { get; }
```
cgranade, line 26: Since this can be null, I'd suggest turning on `#nullable enable` and modifying the type of this field to denote nullability:

```suggestion
            private QuantumSimulator? Simulator { get; }
```

irinayat-MS:
It looks like `TrotterEstimateEnergy` in https://github.com/microsoft/QuantumLibraries/blob/main/Chemistry/tests/JupyterTests/sample.py is attempting to prepare state on qubits that aren't in state |0...0>.

@cgranade:
1. Documentation for PrepareArbitraryState[https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.preparation.preparearbitrarystate?view=qsharp-preview] says "Qubit register encoding number states in little-endian format. This is _expected_ to be initialized in the computational basis state |0...0⟩.". I had interpreted "expected" as a hard prerequisite and made Quantum Simulator throw, when the condition is violated. However, it seems that the current implementation of the API doesn't fail on non-zero qubits -- is it a bug or "expected" means that we simply don't guarantee a meaningful result if the target qubits aren't in zero state?

2. If throwing is the right thing to do, who owns this test and could help me to fix it to satisfy the prerequisites?

cgranade (commented):
Not had a chance to do a complete review yet, sorry; please treat these comments as provisional for now. Thanks for your patience!

cgranade:
> It looks like `TrotterEstimateEnergy` in https://github.com/microsoft/QuantumLibraries/blob/main/Chemistry/tests/JupyterTests/sample.py is attempting to prepare state on qubits that aren't in state |0...0>.
> 
> @cgranade:
> 
> 1. Documentation for [PrepareArbitraryState](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.preparation.preparearbitrarystate?view=qsharp-preview) says "Qubit register encoding number states in little-endian format. This is _expected_ to be initialized in the computational basis state |0...0⟩.". I had interpreted "expected" as a hard prerequisite and made Quantum Simulator throw, when the condition is violated. However, it seems that the current implementation of the API doesn't fail on non-zero qubits -- is it a bug or "expected" means that we simply don't guarantee a meaningful result if the target qubits aren't in zero state?

Since `PrepareAribtraryState` supports `is Adj + Ctl`, it has to do satisfy the invariant that `PrepareArbitraryState(input); Adjoint PrepareArbitraryState(input);` is a no-op even if the qubits in `input` are not in the |00…0⟩ state initially. From that perspective, the verb "Prepare" is used to communicate that the action of the operation is fully defined only for |00…0⟩ inputs, but is guaranteed to at least be unitary for other inputs.

My suggestion would thus be to add a check similar to the `CanEmulate` method in `EstimateFrequencyA` that only uses the emulation feature if the register starts off in the all-zeros state and is large enough to get some advantage from the emulation feature.

Standard/src/Preparation/Arbitrary.cs
```diff
@@ -0,0 +1,66 @@
...
+                this.Simulator = m as QuantumSimulator;
+            }
+
+            /// <summary>
+            /// Overrides the body to do the emulation when possible. If emulation is not possible, then
+            /// it just invokes the default Q# implementation.
+            /// </summary>
+            public override Func<(IQArray<Math.ComplexPolar>, Arithmetic.LittleEndian), QVoid>__Body__ => (_args) =>
+            {
+                var (polar_amplitudes, qubits) = _args;
```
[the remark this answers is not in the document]
irinayat-MS: Sorry, haven't switched the coding style after working in the native simulator :)

Standard/src/Preparation/Arbitrary.cs
irinayat-MS: > My suggestion would thus be to add a check similar to the `CanEmulate` method in `EstimateFrequencyA` that only uses the emulation feature if the register starts off in the all-zeros state and is large enough to get some advantage from the emulation feature.

In general we cannot check the register for being in zero state, can we? I can change the API on the native side to return bool and return `false` without modifying the state, if the prerequisites aren't met. The caller then can fallback to Q# implementation. The disadvantage of this approach is that we'll unpack amplitudes even if we end up not using them. Another way is to add something like `bool CanInjectState(qubits)` to the public C-API surface of the native simulator and call it before calling `InjectState`. In some sense, it's cleaner, but I'm hesitant to grow the C-API surface of the simulator too gratuitously... I think, I'll go with the first of the two solutions.

irinayat-MS:
> My suggestion would thus be to add a check similar to the `CanEmulate` method in `EstimateFrequencyA` that only uses the emulation feature if the register starts off in the all-zeros state and is large enough to get some advantage from the emulation feature.

In general we cannot check the register for being in zero state, can we? I can change the API on the native side to return bool and return `false` without modifying the state, if the prerequisites aren't met. The caller then can fallback to Q# implementation. The disadvantage of this approach is that we'll unpack amplitudes even if we end up not using them. Another way is to add something like `bool CanInjectState(qubits)` to the public C-API surface of the native simulator and call it before calling `InjectState`. In some sense, it's cleaner, but I'm hesitant to grow the C-API surface of the simulator too gratuitously... I think, I'll go with the first of the two solutions.

Standard/src/Preparation/Arbitrary.cs
irinayat-MS: The injected state is precise so the difference will depend on the library's implementation of the prepare gate sequence and the tolerance specified by the user.

cgranade:
> In general we cannot check the register for being in zero state, can we?

I think there's already a native API for that, used to implement `AssertMeasurement` and `AssertMeasurementProbability`?

```C++
    double JointEnsembleProbability(std::vector<Gates::Basis> bs, std::vector<logical_qubit_id> qs)
    {
        removeIdentities(bs, qs);
        if (bs.empty())
        {
            return 0.0;
        }

        recursive_lock_type l(mutex());
        changebasis(bs, qs, true);
        double p = psi.jointprobability(qs);
        changebasis(bs, qs, false);
        return p;
    }
```

https://github.com/microsoft/qsharp-runtime/blob/1224db00db42333316692f8242129101292ab4e5/src/Simulation/Native/src/simulator/simulator.hpp#L56

> I can change the API on the native side to return bool and return `false` without modifying the state, if the prerequisites aren't met. The caller then can fallback to Q# implementation. The disadvantage of this approach is that we'll unpack amplitudes even if we end up not using them. 

That makes a lot of sense, yeah. Hopefully should be able to see performance implications of that unpacking in end-to-end benchmarks. 

> Another way is to add something like `bool CanInjectState(qubits)` to the public C-API surface of the native simulator and call it before calling `InjectState`. In some sense, it's cleaner, but I'm hesitant to grow the C-API surface of the simulator too gratuitously... I think, I'll go with the first of the two solutions.

Makes sense, agreed; if it turns out from benchmarks that we need to grow the C-API surface further, maybe we could generalize slightly to allow checking registers of qubits against |00…0⟩, then do the check from the C# side?

> Note: Before I update the runtime not to throw, added a catch so we can run all tests and discover other potential problems. This is not meant to be final.

Sounds good!

irinayat-MS:
> I think there's already a native API for that, used to implement `AssertMeasurement` and `AssertMeasurementProbability`?

I meant to say, that there is no generic Q# way to check the state of the register because only specific simulators can do this, for example the full state simulator. And as long as we have to ping the simulator anyway, we can do the check inside InjectState API or in an additional call as I described above. Also, to guarantee the prereq, AssertMeasurementProbability has to be done on each qubit separately, right? So if we do it on Q# side it will have to make n roundtrips to the simulator rather than a single call if we do the check inside InjectState.

Note: Considered not doing the check at all and allowing to inject the state anyway. Unfortunately, in the current implementation of partial state injection this might not end up being a unitary operation so cannot go that route.

> Hopefully should be able to see performance implications of that unpacking in end-to-end benchmarks.

Even if we do see the impact, I'm not sure it's worth optimizing here, because the result of the behaviour is going to be undefined anyway, no?

However, there is one more concern I have now. If I understand your previous explanations correctly, we must provide precise emulation of adjoint as well, otherwise base*adjoint might not yield identity.

bettinaheim (approved):

Standard/src/Preparation/Arbitrary.cs
```diff
@@ -0,0 +1,82 @@
...
+            /// <summary>
+            /// Overrides the body to do the emulation when possible. If emulation is not possible, then
+            /// it just invokes the default Q# implementation.
+            /// </summary>
+            public override Func<(IQArray<Math.ComplexPolar>, Arithmetic.LittleEndian), QVoid>__Body__ => (_args) =>
+            {
+                var (polarAmplitudes, qubits) = _args;
+
+                // TODO: benchmark for small `qubits` arrays to find out in which cases emulation is actually
+                // benefitial.
```
cgranade, line 43:
```suggestion
                // beneficial.
```

Standard/src/Preparation/Arbitrary.qs
```diff
@@ -161,6 +161,21 @@ namespace Microsoft.Quantum.Preparation {
         ApproximatelyPrepareArbitraryState(0.0, coefficients, qubits);
     }
 
+    /// # Summary
+    /// Given a set of coefficients and a little-endian encoded quantum register
```
cgranade, line 165: Would you be willing to add a quick note to the summary indicating that this is for internal use only? I'd love if we could use the `internal` keyword directly, but unfortunately we'll need to call this across assembly boundaries...

cgranade (approved):
I think this looks good, thanks for doing this! Note that we won't see improvements in QML performance until we port that library to use `_PrepareAmplitudesFromZeroState`.

Standard/src/Preparation/Arbitrary.qs
irinayat-MS: I'm not sure about where the internal versus public line goes here. Aren't we adding this specifically to allow the client code to choose whether to enable emulation or not?

cgranade: Yes, but on an internal-only basis while we get the right public Q# API designed and reviewed (possibly using the initializer proposal at https://github.com/microsoft/qsharp-language/pull/41).

PR #371, opened 2020-11-12
Add Psi4 and OpenMolcas codegen functionality

guenp:
Add functions to generate Psi4 and OpenMolcas input format

### DIFF
--- a/Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
+++ b/Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
@@ -36,7 +36,7 @@ Group={symmetry}
 &SCF
 Charge={charge}
 Spin={spin}
-{rasscf}
+{rasscf_section}
 """
 
 OPENMOLCAS_TEMPLATE_CASSCF = """
@@ -56,8 +56,8 @@ OPEN_MOLCAS_TEMPLATE_FCIDUMP = """
   Charge={charge}
 
   RGINPUT
-  nsweeps = {nsweeps}
-  max_bond_dimension = {max_bond_dimension}
+  nsweeps = 5
+  max_bond_dimension = 500
   ENDRG
 """
 
@@ -71,12 +71,11 @@ def create_input_deck(
     basis: str = "ANO-RCC-MB",
     symmetry: str = "C1",
     integral_keyword: str = "",
-    method: str = "",
+    method: str = "CASSCF",
+    get_broombridge: bool = False,
     num_active_el: int = None,
     num_active_orbitals: int = 0,
-    ci_root: int = 1,
-    nsweeps: int = 5,
-    max_bond_dimension: int = 500
+    ci_root: int = 1
 ) -> str:
     """Convenience function for creating an OpenMolcas input formatted string.
 
@@ -94,20 +93,18 @@ def create_input_deck(
     :type basis: str, optional
     :param symmetry: Molecule summetry, defaults to "C1"
     :type symmetry: str, optional
-    :param integral_keyword: Integral keyword, e.g. "Cholesky", defaults to ""
+    :param integral_keyword: Integral keyword, e.g. "Cholesky", "DIRECT", "1CCD", defaults to ""
     :type integral_keyword: str, optional
     :param method: Method to use, e.g. "FCIDUMP" or "CASSCF", defaults to ""
-    :type method: str, optional
+    :type method: str, defaults to "CASSCF"
+    :param get_broombridge: Get data for generating Broombridge file
+    :type get_broombridge: bool, defaults to False
     :param num_active_el: Number of active electrons, defaults to None (will be calculated from molecule if unspecified)
     :type num_active_el: int, optional
     :param num_active_orbitals: number of orbitals in each symmetry for the RAS2 orbital subspace, defaults to 0
     :type num_active_orbitals: int, optional
     :param ci_root: CI root(s), see https://molcas.gitlab.io/OpenMolcas/sphinx/users.guide/programs/rasscf.html#optional-keywords, defaults to 1 (ground state)
     :type ci_root: int, optional
-    :param nsweeps: Number of sweeps for FCIDUMP method, defaults to 5
-    :type nsweeps: int, optional
-    :param max_bond_dimension: Maximum bond dimension for FCIDUMP method, defaults to 500
-    :type max_bond_dimension: int, optional
     :return: OpenMolcas input string
     :rtype: str
     """
@@ -119,25 +116,23 @@ def create_input_deck(
     num_atoms = len(geometry) if isinstance(geometry, Geometry) else num_atoms_from_mol(mol)
     charge = charge if charge is not None else GetFormalCharge(mol)
 
-    if method.upper() == "CASSCF":
+    if method.upper() == "CASSCF" and get_broombridge is False:
         num_active_el = formatted_num_active_el(mol=mol, num_active_el=num_active_el)
-        rasscf = OPENMOLCAS_TEMPLATE_CASSCF.format(
+        rasscf_section = OPENMOLCAS_TEMPLATE_CASSCF.format(
             charge=charge,
             num_active_el=num_active_el,
             num_active_orbitals=num_active_orbitals,
             nroot=ci_root
         )
 
-    elif method.upper() == "FCIDUMP":
-        rasscf = OPEN_MOLCAS_TEMPLATE_FCIDUMP.format(
+    elif get_broombridge is True:
+        rasscf_section = OPEN_MOLCAS_TEMPLATE_FCIDUMP.format(
             spin=spin,
-            charge=charge,
-            nsweeps=nsweeps,
-            max_bond_dimension=max_bond_dimension
+            charge=charge
         )
 
     else:
-      rasscf = ""
+      rasscf_section = ""
 
     return OPENMOLCAS_TEMPLATE.format(
         num_atoms=num_atoms,
@@ -148,5 +143,5 @@ def create_input_deck(
         integral_keyword=integral_keyword,
         charge=charge,
         spin=spin,
-        rasscf=rasscf
+        rasscf_section=rasscf_section
     )
--- a/Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
+++ b/Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
@@ -4,7 +4,7 @@
 """Module for generating OpenMolcas input deck format.
 """
 
-from typing import Union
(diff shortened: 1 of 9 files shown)


Python/qdk-chemistry/qdk_chemistry/solvers/psi4.py
```diff
@@ -0,0 +1,75 @@
...
+    method: str = "SCF",
+    driver: str = "energy",
+    scf_type: str = "pk",
+    memory_in_gb: int = 1,
+    reference: str = "rhf",
+    num_active_el: int = None,
+    num_active_orbitals: int = 0,
+    ci_root: int = 1,
+    nsweeps: int = 5,
+    max_bond_dimension: int = 500,
```
dgwvfxn: Delete 
    num_active_el: int = None,
    num_active_orbitals: int = 0,
    ci_root: int = 1,
    nsweeps: int = 5,
    max_bond_dimension: int = 500,
as they as OpenMolcas keywords.

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
```diff
@@ -0,0 +1,152 @@
...
+    spin: str = 1,
+    basis: str = "ANO-RCC-MB",
+    symmetry: str = "C1",
+    integral_keyword: str = "",
+    method: str = "",
+    num_active_el: int = None,
+    num_active_orbitals: int = 0,
+    ci_root: int = 1,
+    nsweeps: int = 5,
+    max_bond_dimension: int = 500
```
dgwvfxn, line 79: Let's delete nsweeps and max_bond_dimension, these two are placeholders, so we can hard code the value.

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
```diff
@@ -0,0 +1,152 @@
...
+
+def create_input_deck(
+    mol: "Mol",
+    mol_name: str = "",
+    geometry: Union[str, Geometry] = None,
+    charge: int = None,
+    spin: str = 1,
+    basis: str = "ANO-RCC-MB",
+    symmetry: str = "C1",
+    integral_keyword: str = "",
```
dgwvfxn, line 73: The integral keyword are "Cholesky", "DIRECT", "1CCD", or empty. We should use empty as default.

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
```diff
@@ -0,0 +1,152 @@
...
+Basis={basis}
+Group={symmetry}
+
+&SEWARD
+{integral_keyword}
+
+&SCF
+Charge={charge}
+Spin={spin}
+{rasscf}
```
dgwvfxn, line 39: what is {rasscf} here?

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
```diff
@@ -0,0 +1,152 @@
...
+def create_input_deck(
+    mol: "Mol",
+    mol_name: str = "",
+    geometry: Union[str, Geometry] = None,
+    charge: int = None,
+    spin: str = 1,
+    basis: str = "ANO-RCC-MB",
+    symmetry: str = "C1",
+    integral_keyword: str = "",
+    method: str = "",
```
dgwvfxn, line 74: OpenMolcas is a method specific package, mainly focus on "CASSCF", "RASSCF", "CI" and "PT2" corrections. In this version, let's focus on "CASSCF" method functionality. 
Basically, if the method is "HF", then we stop at &SCF section, if the method is "CASSCF", then we append the &RASSCF section.  
As you may noticed, unlike Psi4 and NWCHEM, we don't have {driver} in OpenMolcas, this is because we currently only focus on energy calculation of CASSCF method.

Python/qdk-chemistry/qdk_chemistry/solvers/psi4.py
```diff
@@ -1,25 +1,75 @@
...
-  d_convergence {1e-8}
-  e_convergence {1e-8}
-}
+  scf_type {scf_type}
+  reference {reference}
+  d_convergence {d_convergence:.1e}
+  e_convergence {e_convergence:.1e}
+}}
+
+{method}
```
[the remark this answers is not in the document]
guenp: Agreed; this is a placeholder for the entire block. But I'll rename it for clarity.

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
guenp: This is a placeholder for the &RASSF section which I save in a separate variable.

dgwvfxn: Ah I see!

Python/qdk-chemistry/qdk_chemistry/solvers/util.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def formatted_num_active_el(mol: "Mol", num_active_el: int = None):
+    """Get the number of active electrons either from the molecule object 
+    or input value.
+
+    :param mol: Molecule object
+    :type mol: Mol, optional
+    :param num_active_el: Number of active electrons
+    :type int, optional
```
rmshaffer, line 66:
```suggestion
    :type num_active_el: int, optional
```

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    formatted_geometry = formatted_geometry_str(
+        mol=mol, 
+        geometry=geometry, 
+        line_sep=GEOMETRY_LINE_SEP
+    )
+    num_atoms = len(geometry) if isinstance(geometry, Geometry) else num_atoms_from_mol(mol)
+    charge = charge if charge is not None else GetFormalCharge(mol)
+
+    if method.upper() == "CASSCF" and get_broombridge is False:
```
rmshaffer, line 119: as per PEP-8, `is True` / `is False` are very undesirable (I don't feel nearly as strongly, but I defer 😊)
```suggestion
    if method.upper() == "CASSCF" and not get_broombridge:
```

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
```diff
@@ -0,0 +1,147 @@
...
+        num_active_el = formatted_num_active_el(mol=mol, num_active_el=num_active_el)
+        rasscf_section = OPENMOLCAS_TEMPLATE_CASSCF.format(
+            charge=charge,
+            num_active_el=num_active_el,
+            num_active_orbitals=num_active_orbitals,
+            nroot=ci_root
+        )
+
+    elif get_broombridge is True:
```
rmshaffer, line 128:
```suggestion
    elif get_broombridge:
```

rmshaffer (approved):
I'm not the best person to review this for functionality, but syntax-wise and structure-wise I think it looks good!

Python/qdk-chemistry/qdk_chemistry/solvers/openmolcas.py
guenp: I prefer to check explicitly if this is set to `True` in case someone accidentally passes the wrong value.

rmshaffer: Yeah, I can see that. Something I learned is that `is True` determines if something is literally the Python built-in `True`, but if someone passes in an object of a different boolean type such as a `numpy.bool_`, then `is True` will always be false. I am not sure it's a big deal here either way, but I think the pythonic way is to just omit the `is True`. But then I don't know the best way to do input type checking.

guenp: Yes I agree, `if foo` is the [PEP8 preferred way](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html), in this case I prefer to edge on the side of caution because this function generates a payload to be sent to an API that could take a long time to run so I want to make sure it only adds the Broombridge part if the user explicitly specifies `get_broombridge=True` and not if they accidentally pass `"foo"` as a positional argument in the wrong spot.
Speaking of type checking; I believe @cgranade floated the idea to support `mypy` at some point in the future so we can start using [Parameterizing Literals](https://mypy.readthedocs.io/en/stable/literal_types.html#parameterizing-literals) as an alternative to enums

cgranade: If I understand correctly, PEP8 says that _either_ what you currently have `if get_broombridge is True` or @rmshaffer's suggestion of `if get_broombridge` is preferred, depending on the exact condition that needs to be checked. The bad case to avoid is `if get_broombridge == True`, since that has neither the brevity of the first form nor the explicitness of the second case. My _personal_ preference tends towards `if cond:` following "better to ask forgiveness than permission" thinking that's idiomatic to Python, but I totally take the point that asking forgiveness before launching a longer-running computation may not be ideal.

In any case, the two are equivalent if the caller and the library are both correctly guarded by `mypy`; that would be a longer project, but as the code volume we have in Python increases, may well be worth it at some point.

rmshaffer: Yes, I agree with everything you're both saying, and I think for the reasons @guenp points out it's totally reasonable to leave the `is True` check.

PEP 8 seems to quite explicitly say that `is True` is _worse_ than `== true`, but I can't really make sense of that given the totally valid need to type-check in a case like this. What do they want? `if isinstance(get_broombridge, bool) and get_broombridge:` 😂

guenp: > Don't compare boolean values to True or False using ==.
> 
> Yes:   if greeting:
> No:    if greeting == True:
> Worse: if greeting is True:

@rmshaffer oops you're right, I must've misread...
But yes it looks like `if instance(get_broombridge, bool) and get_broombridge` is indeed the PEP8-preferred option 🤷

cgranade: I must admit, I didn't expect that monstrosity to be the PEP8 recommendation.... seems like a perfectly reasonable place to make an exception from PEP8? I get it's more explicit than `is True`, but I think the meaning of `is` is perfectly clear to any Python developers that would be maintaining the codebase in the future......

msoeken (approved):
Thanks! The implementation looks good to me. Cannot judge much on the semantic details of the code generation though.

PR #378, opened 2020-11-19
Minor improvements to C# code quality.

cgranade:


anjbur (commented):
In PauliTerm.cs, the call to `ReferenceEquals` on line 178 (formerly line 199) was also flagged. Otherwise looks good to me!

cgranade:
Thanks for the review, @anjbur! I think that should now be fixed. There was also a build failure caused by not implementing `IEquatable<T>` correctly, so I went on and fixed that as well by using a pattern match on `IEquatable<T>` from `object.Equals`. Thanks again!

Standard/src/Emulation/Math.cs
```diff
@@ -192,9 +194,9 @@ private long SampleUsingBTPE()
             if (v > F) goto Step10;
             goto Step60;
 
-            Step52:
+        Step52:
             rho = (k/(nrq))*((k*(k/3.0 + 0.625) + 0.16666666666666666)/nrq + 0.5);
-            t = -k*k/(2*nrq);
+            t = -k*k/(2*nrq); // lgtm [cs/loss-of-precision]
```
guenp, line 199: Not sure I understand this comment?

cgranade: It's an instruction to CI to indicate that a warning raised about that line is a false positive.

msoeken (approved):

anjbur (approved):
Looks good to me, thank you!

anpaz (approved):
lgtm.

Issue #113, opened 2019-05-24
Provide more examples in API documentation comments


cgranade:
In general, the API documentation pages for the Q# libraries are fairly light on examples. Adding more `/// # Example` sections would help make the libraries more accessible and easier to use.

guenp:
This is a good idea but the issue is a bit too general for us to actually get started on this.
My suggestions for action items:

- [ ] Create a new issue to add examples to one or two frequently used namespaces
- [ ] Explicitly make examples part of the Q# Libraries API proposal template.
- [ ] Close this issue

Tagging @cgranade @msoeken

guenp:
Interesting find when @msoeken were looking into the docs: https://github.com/MicrosoftDocs/quantum-docs-pr/issues/1121

msoeken:
Closing as duplicate to #339

PR #381, opened 2020-11-23
Add new issue template for design proposals.

cgranade:


.github/ISSUE_TEMPLATE/api_design_proposal.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Modifications to style guide
+
+<!-- Does your proposal require modifying the style guide or API design principles? If so, please list and discuss any required modifications here, as that carries a unique impact that requires separate discussion. -->
+
+### Impact of breaking changes
+
+<!-- If your proposal includes breaking changes beyond @Deprecated attributes, please list those changes here and their expected impact on Q# users. -->
+
+## Example
```
guenp, line 42:
```suggestion
## Examples
```

.github/ISSUE_TEMPLATE/api_design_proposal.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Current status
+
+<!-- As appropriate, list examples of using the current Q# API. -->
+
+### Using proposed changes
+
+<!-- As appropriate, list examples of using the changes included in your proposal. -->
```
guenp, line 50:
```suggestion
<!-- As appropriate, list examples of using the desired Q# API for each of the functions, operations and UDTs you are introducing or changing. When implementing said features, please make sure to include these examples in the docstring using the "/// # Example" formatting, such that the examples will be visible in the documentation. -->
```

guenp (approved):
Thanks, @cgranade for adding this template! Looks good to me. I added some suggestions based on a discussion with @msoeken this morning (see task added on issue https://github.com/microsoft/QuantumLibraries/issues/113).

.github/ISSUE_TEMPLATE/api_design_proposal.md
cgranade: While I agree it's important to stress writing good API docs, I think that's more of an issue with the implementation of an API and not the design itself? Given that, I'd prefer to keep that kind of advice in our contributing docs at https://docs.microsoft.com/en-us/quantum/resources/contributing/ so as to keep this part focused more tightly on the template needed for a design proposal itself.

guenp: Ok, in that case let's discuss in our API meeting and make sure that we catch this in the code review process.

msoeken: I suggest to add `/// # Example` stubs to each entry under _New and modified functions, operations, and UDTs_

msoeken:
Thanks @cgranade for setting up this template. I think that this template may be good for someone very familiar with the QDK libraries and the API review process, but might be overwhelming for a newcomer. Could we make more sections optional that we can fill in later during discussion or by ourselves. Or can we add some comment at the top of the proposal saying that it's okay to leave some sections empty if they are unclear. I am just a little concerned to discourage involvement because of the process.

msoeken (approved):

.github/ISSUE_TEMPLATE/api_design_proposal.md
cgranade: Done!

PR #397, opened 2021-01-13
Slight improvements to error handling in TestPurifiedMixedState.

cgranade:
This PR slightly improves the `TestPurifiedMixedState` unit test to help diagnose failures such as that seen in https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=21163&view=logs&j=b78a1ffd-6a4f-5c9a-fd79-c70d65470011&t=a3b90293-a206-53fd-a1c1-8b7b4b27ab86.

### DIFF
--- a/Standard/tests/QuantumROMTests.qs
+++ b/Standard/tests/QuantumROMTests.qs
@@ -65,8 +65,9 @@ namespace Microsoft.Quantum.Tests {
 
     @Test("QuantumSimulator")
     operation TestPurifiedMixedState() : Unit {
-        for(coeffs in 2..7){
-            for(nBitsPrecision in -1..-1..-2){
+        for (coeffs in 2..7) {
+            for (nBitsPrecision in -1..-1..-2) {
+                Message($"[TestPurifiedMixedState] Test case: coeffs = {coeffs}, nBitsPrecision = {nBitsPrecision}");
                 let targetError = PowD(2.0, IntAsDouble(nBitsPrecision));
                 let probtargetError = targetError / IntAsDouble(coeffs);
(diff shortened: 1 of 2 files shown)


Standard/tests/QuantumROMTests.qs
```diff
@@ -65,8 +65,9 @@ namespace Microsoft.Quantum.Tests {
 
     @Test("QuantumSimulator")
     operation TestPurifiedMixedState() : Unit {
-        for(coeffs in 2..7){
-            for(nBitsPrecision in -1..-1..-2){
+        for (coeffs in 2..7) {
+            for (nBitsPrecision in -1..-1..-2) {
+                Message($"[TestPurifiedMixedState] Test case: coeffs = {coeffs}, nBitsPrecision = {nBitsPrecision}");
```
msoeken, line 70: I would prefer to only output text when tests go wrong. If everything goes right, I think it's nicer that the tests are silent.

msoeken (approved):
LGTM, I just left one comment. Thanks!

Standard/tests/QuantumROMTests.qs
cgranade: Makes sense, and I tend to agree more generally. I only put that in in this case, since the test loops over an unusually large number of cases, making failures such as the one in build 21163 harder to diagnose. Are you OK with this as an exception?

msoeken: Okay, that makes sense. Thanks for the additional information.

cgranade: No worries, thanks for catching that!

PR #401, opened 2021-01-21
Adding explicit reference to Qsharp.Core package

swernli:
With the ongoing refactor of the qsharp-runtime repo to support target packages, the Microsoft.Quantum.Simulators package no longer automatically brings in the Microsoft.Quantum.QSharp.Core package, so projects need to explicitly add that reference if they are have suppressed loading of the core package via `IncludeQsharpCorePackages` set to false. See microsoft/qsharp-runtime#476, which this change will help unblock.


cgranade:
Looks like the build failed with a transient error (http 503 getting NuGet packages), so re-ran failed jobs.

cgranade (approved):

PR #404, opened 2021-02-01
Iterative BitSizeL implementation

### LINKED ISSUE #403: Stack overflow when using `Microsoft.Quantum.Math.BitSizeL`
**Describe the bug**
Stack overflow when using `Microsoft.Quantum.Math.BitSizeL` with moderate size big integers.

**To Reproduce**
Create project consisting of single Q# file:
```qsharp 
namespace Stackoverflow {

    open  Microsoft.Quantum.Math;

    @EntryPoint()
    operation HelloQ() : Unit
    {
        for k in 1 .. 2048 {
            Message($"Trying 1L <<< {128*k}");
            let b = BitSizeL(1L <<< (128*k));
            Message($"BitSizeL(1 <<< {128*k}) is {b}");
        }
    }
}
```
Output : 
```
...
BitSizeL(1 <<< 2688) is 2689
Trying 1L <<< 2816
BitSizeL(1 <<< 2816) is 2817
Trying 1L <<< 2944
Stack overflow.
```

**Expected behavior**
Q# BigInt type has similar performance and uses to C# BigInt type.

**System information**
 - OS: Windows
 - `Microsoft.Quantum.Sdk/0.15.2101125897`

**Additional context** 
Try the following C# code 
```csharp
using System;
using System.Numerics;

namespace bigint
{
    class Program
    {
        static void Main(string[] args)
        {
            for( int k = 0; k < 31; ++k )
            {
                Console.WriteLine($"Trying 1 << {1 << k}");
                long b = ((new BigInteger(1)) << (1 << k)).GetBitLength();
                Console.WriteLine($"Bitsize of 1 << {(1 << k)} is {b}");
            }
        }
    }
}
```
that runs successfully and outputs 
```
.... 
Trying 1 << 67108864
Bitsize of 1 << 67108864 is 67108865
Trying 1 << 134217728
Bitsize of 1 << 134217728 is 134217729
Trying 1 << 268435456
Bitsize of 1 << 268435456 is 268435457
Trying 1 << 536870912
Bitsize of 1 << 536870912 is 536870913
Trying 1 << 1073741824
Bitsize of 1 << 1073741824 is 1073741825
```

msoeken:
This PR fixes #403 by replacing the recursive implementation with an iterative one. However, reading @vadym-kl's issue description I wonder whether it is better to implement it as a Q# intrinsic and calling directly into C#'s method to retrieve the bit size.

### DIFF
--- a/Standard/src/Math/Functions.qs
+++ b/Standard/src/Math/Functions.qs
@@ -551,12 +551,6 @@ namespace Microsoft.Quantum.Math {
         return AccumulatedBitsizeI(a, 0);
     }
 
-    /// # Summary
-    /// Helper function used to recursively calculate the bitsize of a value.
-    internal function AccumulatedBitsizeL(val : BigInt, bitsize : Int) : Int {
-        return val == 0L ? bitsize | AccumulatedBitsizeL(val / 2L, bitsize + 1);
-    }
-
 
     /// # Summary
     /// For a non-negative integer `a`, returns the number of bits required to represent `a`.
@@ -572,7 +566,13 @@ namespace Microsoft.Quantum.Math {
     /// The bit-size of `a`.
     function BitSizeL(a : BigInt) : Int {
         Fact(a >= 0L, $"`a` must be non-negative");
-        return AccumulatedBitsizeL(a, 0);
+        mutable bitsize = 0;
+        mutable val = a;
+        while (val != 0L) {
+            set bitsize += 1;
+            set val /= 2L;
+        } 
+        return bitsize;
     }
 
 
--- a/Standard/tests/Math/MathTests.qs
+++ b/Standard/tests/Math/MathTests.qs
@@ -68,6 +68,14 @@ namespace Microsoft.Quantum.Canon {
         EqualityFactI(BitSizeI(3), 2, $"BitSizeI(3) must be 2");
         EqualityFactI(BitSizeI(7), 3, $"BitSizeI(7) must be 2");
     }
+
+    @Test("QuantumSimulator")
+    function CanComputeBitSizeFromLargeNumbers () : Unit {
+        for (k in 1 .. 100) {
+            let exp = 128 * k;
+            Fact(BitSizeL(1L <<< exp) == exp + 1, $"unexpected bitsize for exponent {exp} (k = {k})");
+        }
+    }
     
     @Test("QuantumSimulator")
     function ExpModIsCorrect () : Unit {
--- a/Standard/tests/Math/MathTests.qs
+++ b/Standard/tests/Math/MathTests.qs
@@ -73,7 +73,7 @@ namespace Microsoft.Quantum.Canon {
     function CanComputeBitSizeFromLargeNumbers () : Unit {
         for (k in 1 .. 100) {
             let exp = 128 * k;
-            Fact(BitSizeL(1L <<< exp) == exp + 1, $"unexpected bitsize for exponent {exp} (k = {k})");
+            EqualityFactI(BitSizeL(1L <<< exp), exp + 1, $"unexpected bitsize for exponent {exp} (k = {k})");
         }
     }
     


cgranade:
It looks like the build failure is transient, so maybe let's kick off another one?

/azp run

With respect to the other point, I think especially for QIR emission, it would make sense to keep this backed by your  Q# implementation so as to not introduce another intrinsic that target packages need to implement.

Standard/tests/Math/MathTests.qs
```diff
@@ -68,6 +68,14 @@ namespace Microsoft.Quantum.Canon {
         EqualityFactI(BitSizeI(3), 2, $"BitSizeI(3) must be 2");
         EqualityFactI(BitSizeI(7), 3, $"BitSizeI(7) must be 2");
     }
+
+    @Test("QuantumSimulator")
+    function CanComputeBitSizeFromLargeNumbers () : Unit {
+        for (k in 1 .. 100) {
```
cgranade, line 74:
```suggestion
        for k in 1 .. 100 {
```

Standard/tests/Math/MathTests.qs
```diff
@@ -68,6 +68,14 @@ namespace Microsoft.Quantum.Canon {
...
+            let exp = 128 * k;
+            Fact(BitSizeL(1L <<< exp) == exp + 1, $"unexpected bitsize for exponent {exp} (k = {k})");
```
cgranade, line 76: This will give more informative error messages on failure, as `EqualityFact*` prints out both the actual and expected, while `Fact` can only see if its input was `true` or `false` and not _why_ it got a `false`.
```suggestion
            EqualityFactI(BitSizeL(1L <<< exp), exp + 1, $"unexpected bitsize for exponent {exp} (k = {k})");
```

Standard/src/Math/Functions.qs
```diff
@@ -572,7 +566,13 @@ namespace Microsoft.Quantum.Math {
     /// The bit-size of `a`.
     function BitSizeL(a : BigInt) : Int {
         Fact(a >= 0L, $"`a` must be non-negative");
-        return AccumulatedBitsizeL(a, 0);
+        mutable bitsize = 0;
+        mutable val = a;
+        while (val != 0L) {
```
cgranade, line 571:
```suggestion
        while val != 0L {
```

cgranade (approved):

Standard/tests/Math/MathTests.qs
msoeken: I think the current libraries still builds against an older version of the QDK.

Standard/src/Math/Functions.qs
msoeken: I think the current libraries still builds against an older version of the QDK.

Standard/tests/Math/MathTests.qs
msoeken: That makes sense, thanks. Somehow I was thinking at that moment that `BitSizeL` returns `BigInt` and thought I could not compare.

msoeken:
> With respect to the other point, I think especially for QIR emission, it would make sense to keep this backed by your Q# implementation so as to not introduce another intrinsic that target packages need to implement.

Can we do both? Provide a Q# implementation but also have a C# implementation for when we run Q# applications inside C#? I think such an approach would be generally helpful.

cgranade:
> Can we do both? Provide a Q# implementation but also have a C# implementation for when we run Q# applications inside C#? I think such an approach would be generally helpful.

Good question, yeah. @swernli, I know you've raised that some of the behavior around that has changed recently with the  move to targeting packages; how would be best to have the best of both worlds here?

swernli:
Thanks for drawing my attention to this, Chris! The good news is that the target package work (which merged to qsharp-runtime/main as of https://github.com/microsoft/qsharp-runtime/pull/476) was able to accomplish its goals in a (mostly) non-breaking way. So the behavior I raised before that changed actually didn't change and can still work as it did before. For clarity, here's a list of the current behaviors you can use:

- If you provide a Q# implementation for a callable and a C# implementation, the C# implementation will override the Q# as it does today, using the existing mechanisms for discovering child classes in the C# and registering them as overrides their parent (this behavior was going to change but now has not).
- If you define a Q# callable as `body intrinsic` then by default it will be generated into an abstract C# class, so failing to provide a C# child class with implementation will result in a runtime failure because the infrastructure won't find any overrides and will try to instantiate the abstract class (also matches current behavior).
- If you define a Q# callable as `body intrinsic` and opt-in to the the new behavior it will be generated as a concrete C# class that calls into an interface to invoke the desired implementation. You must then define the interface implement it on your desired simulator or other runtime environment to get the right functionality, and will get compile time errors if the interface definition is missing and runtime errors if the chosen simulator doesn't implement that interface (this is the new behavior and is entirely optional).

So it sounds like in this case you want to use that first option, which will still work as it does today. That way the Q# implementation will be used when generating code (such as for QIR) but the C# implementation will take precedence with our default simulator.

cgranade:
Thanks for the help, @swernli!

> * If you provide a Q# implementation for a callable and a C# implementation, the C# implementation will override the Q# as it does today, using the existing mechanisms for discovering child classes in the C# and registering them as overrides their parent (this behavior was going to change but now has not).
>
> So it sounds like in this case you want to use that first option, which will still work as it does today. That way the Q# implementation will be used when generating code (such as for QIR) but the C# implementation will take precedence with our default simulator.

Agreed, I think this meets both usecases fairly nicely. It's good to have the version that relies on the nicely optimized code from the C# standard libraries while still being able to use the pure Q# version in QIR. My only concern at that point would be testing to make sure that the Q# version is correct when the C# implementation is missing; I think that isn't quite covered by `@Test(...)` at the moment, such that we may need a bit more there...

msoeken:
Thanks for the help and feedback @swernli and @cgranade. I am also not clear at the moment, how we can test both implementations. I will merge this PR for now and eventually open another one, adding a C# native implementation when we sorted out the test questions.

PR #415, opened 2021-02-11
Address optimization in ApplyObliviousAmplitudeAmplification

### LINKED ISSUE #414: Optimization in ApplyObliviousAmplitudeAmplification
I propose to optimize the following code in `ApplyObliviousAmplitudeAmplification`:

https://github.com/microsoft/QuantumLibraries/blob/ac2e60b3d858f97d61f862b5532382c31e7d83f5/Standard/src/AmplitudeAmplification/AmplitudeAmplification.qs#L67-L73

Move the `if` condition around the `within / apply` block, since in case the condition is `false` the two operations in the `within / apply` block are unnecessarily executed, although they cancel each other.

msoeken:
Fixes #414 and addresses some deprecations on the way.

### DIFF
--- a/Standard/src/AmplitudeAmplification/AmplitudeAmplification.qs
+++ b/Standard/src/AmplitudeAmplification/AmplitudeAmplification.qs
@@ -65,9 +65,9 @@ namespace Microsoft.Quantum.AmplitudeAmplification {
             }
 
             if (targetPhase != 0.0) {
-            within {
-                signalOracle!(auxiliaryRegister, systemRegister);
-            } apply {
+                within {
+                    signalOracle!(auxiliaryRegister, systemRegister);
+                } apply {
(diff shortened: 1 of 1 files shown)


Standard/src/AmplitudeAmplification/AmplitudeAmplification.qs
```diff
@@ -64,10 +64,10 @@ namespace Microsoft.Quantum.AmplitudeAmplification {
                 );
             }
 
-            within {
-                signalOracle!(auxiliaryRegister, systemRegister);
-            } apply {
-                if (targetPhase != 0.0) {
+            if (targetPhase != 0.0) {
+                within {
+                    signalOracle!(auxiliaryRegister, systemRegister);
```
cgranade, line 69: Ah, that's a good catch... thanks for fixing this!

cgranade (approved):

PR #422, opened 2021-03-02
corrected invalid documentation reference in ApplyXorInPlace

filipw:
It refers to the old operation `InPlaceXorLE` which it deprecates.


cgranade (approved):
Thanks for catching this!

Issue #386, opened 2020-12-02
AllowAtMostNCallsCA non-deterministic failures


Strilanc:
When I create multiple tests using `AllowAtMostNCallsCA`, I get sporadic failures that look like concurrency bugs.

Repro code:

```
    open Microsoft.Quantum.Diagnostics;
    open Microsoft.Quantum.Intrinsic;

    @Test("ResourcesEstimator")
    operation test1() : Unit {
        for (n in 1..100) {
            using (qs = Qubit[n]) {
                within {
                    AllowAtMostNCallsCA(0, CCNOT, "");
                } apply {
                    op(qs[0], qs[1...]);
                }
            }
        }
    }

    @Test("ResourcesEstimator")
    operation test2() : Unit {
        for (n in 1..100) {
            using (qs = Qubit[n]) {
                within {
                    AllowAtMostNCallsCA(0, CCNOT, "");
                } apply {
                    op(qs[0], qs[1...]);
                }
            }
        }
    }

    operation op(control: Qubit, out_target: Qubit[]) : Unit is Adj {
        for (k in 0..Length(out_target)-1) {
            CNOT(control, out_target[k]);
        }
    }
```

# Expected behavior:

Tests pass.

# Actual behavior:

Test fails sometimes, complaining about iterating a list while mutating it or about an empty stack being popped. For example:

```
Starting test execution, please wait...

A total of 1 test files matched the specified pattern.
Call stack was empty when popped:
System.InvalidOperationException: This operation does not apply to an empty instance.
   at System.Collections.Immutable.ImmutableStack`1.Pop()
   at Microsoft.Quantum.Diagnostics.AllowAtMostNCallsCA`2.Native.<>c__DisplayClass5_0.<get___Body__>b__3(ICallable callable, IApplyData data)
Unhandled exception. System.InvalidOperationException: Stack empty.
[...]
Test Run Failed.
Total tests: 2
     Passed: 1
     Failed: 1
```

and sometimes:

```
Unhandled exception. System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
Unhandled exception. System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
 ---> Microsoft.Quantum.Diagnostics.AllowAtMostNCallsCA [...]
```

**System information**
[...]

PR #431, opened 2021-03-22
Loosen restriction on AllowAtMostNCallsCA operation

### LINKED ISSUE #343: AllowAtMostNCallsCA requires restricted operation to have adjoint and controlled
**Describe the bug**
`AllowAtMostNCallsCA` requires its operation parameter to have adjoint and controlled variants defined, so it can't be used with measurements or any operations returning values (or just operations which don't have these variants). The `'TOutput` in the parameter signature `op : ('TInput => 'TOutput is Adj + Ctl)` seems to imply that using `AllowAtMostNCallsCA` should be possible with operations that return something other than `Unit`, and a lot of use cases for this functionality involves such operations.

**To Reproduce**
The following code does not compile:
```qsharp
within {
    AllowAtMostNCallsCA(0, M, "Too many calls to M.");
} apply {
    // Apply the operation
    op(qs);
}
```

**Expected behavior**
The code compiles and prohibits the operation `op` to use measurements.

**System information**
 - QDK 0.12.20100504

msoeken:
This should fix #343, but I am not sure what the best way to solve it is:

1. Add the other 3 variants `{,A,C}` and leave this one as it is
2. Change the operation name to `AllowAtMostNCalls` and deprecate the current one

@cgranade Would we need API review for this PR, and if so, could we go with option 3:

3. Accept PR for March release and implement option 1 or 2 for the next release.


tcNickolas (commented):
We probably want to add a unit test to check that this change indeed allows measurements to work? Something similar to `CheckAllowNCallsTestShouldFail`

cgranade (approved):
Approving following discussion in #343 and in lieu of any objections from @guenp. I'd also suggest taking @tcNickolas' suggestion of adding a unit test to make sure that the generalization in this PR works as intended.

We should discuss at the next API review cycle how to rename in future releases; @msoeken, would you be willing to update #343 to that effect so we can get it scheduled?

tcNickolas:
I took the liberty to add unit tests to cover the change. I have seen an occasional reoccurrence of #386, but I believe it's been observed before so shouldn't be caused by this particular change.

Standard/tests/Diagnostics/AllowTests.qs
```diff
@@ -34,10 +35,43 @@ namespace Microsoft.Quantum.Tests {
         }
     }
 
+    operation CheckAllowNCallsMeasurementsTestShouldFail() : Unit {
+        within {
+            Diag.AllowAtMostNCallsCA(3, Measure, "Too many calls to Measure.");
+        } apply {
+            using (q = Qubit()) {
```
cgranade, line 42: Minor, but it looks like this test uses deprecated syntax? Definitely should still work, so I'm good to bring this in and fix in the next release as part of a more comprehensive code quality improvement.

cgranade (approved):

Standard/tests/Diagnostics/AllowTests.qs
tcNickolas: Yes, this test uses deprecated syntax for uniformity with the rest of the code which still uses the same syntax. I try not to mix meaningful changes and purely syntactic changes, since that makes following the changelog later much more painful.

PR #434, opened 2021-03-24
Fix broken documentation links

anjbur:


tcNickolas (commented):
Standard/README.md shows as binary (weird!), but looking at it in https://github.com/microsoft/QuantumLibraries/blob/2c2c786fd8a75eb0295da2adbd8204cf54e54e23/Standard/README.md, it says "Q# sources used to implement the canon portion" but points to [chemistry](https://docs.microsoft.com/azure/quantum/user-guide/libraries/chemistry) - this doesn't look right?

CONTRIBUTING.md
```diff
@@ -3,7 +3,7 @@
 Welcome, and thank you for your interest in contributing to the Quantum Development Kit!
 
 There are many ways in which you can contribute, beyond writing code. The goal of this document is to provide a high-level overview of how you can get involved.
-For more details on how to contribute to the samples or to the rest of the Quantum Development Kit, please see the [contribution guide](https://docs.microsoft.com/quantum/contributing/).
+For more details on how to contribute to the samples or to the rest of the Quantum Development Kit, please see the [contribution guide](https://docs.microsoft.com/azure/quantum/contributing-overview/).
```
cgranade, line 6: That's weird, I wonder why this was moved?

cgranade (commented):
> Standard/README.md shows as binary (weird!),

That can often be an artifact of some tools not correctly parsing byte order marks in UTF-8 encoded files (UTF-8 with BOM is a non-standard MS extension to the Unicode spec). As a result, we may be able to fix by removing the first two bytes. Using `Set-Content -Encoding Utf8NoBOM` to normalize the README confirms that this is likely the case.

CONTRIBUTING.md
anjbur: Unclear, it looks like it should point to the same place though! It links to the main page in the "Contributing to the QDK" documentation section

anjbur:
The link is fixed. The README still shows as a binary in the "Files Changed" summary even though it correctly looks like a text file if you look at just the most recent commit. @cgranade is that something that needs to be fixed? Do you know why it would show differently in the per-commit view versus the overall view?

tcNickolas (approved):

PR #439, opened 2021-04-07
Add Factorial function within the standard math library.

TheMagicNacho:
**Issue**
The standard math library does not have a function for Factorials.

**Discussion**
Factorials are commonly used functions in math and science packages, Q# did not have a factorial function previously.

**Recommendation**
I created a Factorial Function which returns an Int type. 
I recommend to merge as this function does not impede with other functionalities.

P.S. I'm also working on a function which returns a BigInt.


TheMagicNacho:
I added committed an additional function which takes an Int and returns a BigInt.

TheMagicNacho:
Closing to fix some build errors.

PR #440, opened 2021-04-07
Add factorial functions for Int and BigInt.

TheMagicNacho:
**Issue**
The Q# standard libraries do not have a factorial function

**Discussion**
Factorials are a commonly used function within physics and probability.

**Recommendation**
I have created two functions which return the factorial of a given Int type.
`FactorialI(n)` Returns an Int so long the number is within the domain [-20,20].
`FactorialB(n)` Returns a BigInt type given any Int.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ packages/
 *.exe
 *.chm
 
+# Pipeline drops dir
+drops/
+
 # test outputs
 TestResults/
 *.qxe
--- a/.gitignore
+++ b/.gitignore
@@ -335,3 +335,5 @@ __pycache__/
 
 /ionqTest
 /qsharpTest
+/ionQTest-2
+/ionQTest-2.csproj
--- a/.gitignore
+++ b/.gitignore
@@ -330,5 +330,5 @@ __pycache__/
 *.odx.cs
 *.xsd.cs
 
-ionQTest-2/Program.qs
-ionQTest-2/ionQTest-2.csproj
+/ionqTest
+/qsharpTest
--- a/.gitignore
+++ b/.gitignore
@@ -332,8 +332,3 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
-
-/ionqTest
-/qsharpTest
-/ionQTest-2
-/ionQTest-2.csproj
--- a/.gitignore
+++ b/.gitignore
@@ -330,3 +330,5 @@ __pycache__/
 *.odx.cs
 *.xsd.cs
 
+ionQTest-2/Program.qs
+ionQTest-2/ionQTest-2.csproj
--- a/Standard/src/Math/Functions.qs
+++ b/Standard/src/Math/Functions.qs
@@ -681,13 +681,11 @@ namespace Microsoft.Quantum.Math {
             set an = AbsI(n);
             set x = -1;
         }
-        elif(n == 0){
+        elif n == 0 {
             return x;
-        }
-        elif(n >= 21){
-            fail "Largest factorial an Int can hold is |20|. Use Factorial BigInt.  \n \n >FactorialB(n) .";
-        }
-        else{
+        } elif n >= 21 {
+            fail "Largest factorial an Int can hold is 20!. Use Factorial BigInt.  \n \n >FactorialB(n).";
+        } else {
             set an = n;
         }
         for i in  1 .. an {
--- a/Standard/src/Math/Functions.qs
+++ b/Standard/src/Math/Functions.qs
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
             return output;
         }
     }
+    
+    /// # Summary
+    /// Returns a factorial as an int.
+    ///
+    /// # Description
+    /// Returns the factorial as Integer, given an input of $n$ as an Integer.
+    /// Maximum input is |20| to return as Int. For inputs greater than 20, use FactorialB(n).
+    ///
+    /// # Input
+    /// ## $n$
+    /// An Int between -20, 20.
+    ///
+    /// # Output
+    /// The factorial of the provided input with the datatype Int.
+    function FactorialI (n : Int) : Int
+    {
+        mutable an = 1;
+        mutable x = 1;
+        if (n < 0 ){
+            set an = AbsI(n);
+            set x = -1;
+        }
+        elif(n == 0){
+            return x;
+        }
+        elif(n >= 21){
+            fail "Largest factorial an Int can hold is |20|. Use Factorial BigInt.  \n \n >FactorialB(n) .";
+        }
+        else{
+            set an = n;
+        }
+        for i in  1 .. an {
+            set x = x * i;
+        }
+       return x;
+    }
+
+    /// # Summary
+    /// Returns a factorial as a BigInt.
+    ///
+    /// # Description
+    /// Returns the factorial as Big Integer, given an input of $n$ as an Integer.
+    ///
+    /// # Input
+    /// ## $n$
+    /// A whole number of any size, positive or negative.
+    ///
+    /// # Output
+    /// The factorial of the provided input with the type BigInt.
+    function FactorialB (n : Int) : BigInt
+    {
+        mutable an = 1;
+        mutable x = IntAsBigInt(1);
+        if (n < 0 ){
+            set an = AbsI(n);
+            set x = IntAsBigInt(-1);
+        }
+        elif(n == 0){
+            return x;
+        }
+        else{
+            set an = n;
+        }
+        for i in  1 .. an {
+            set x = x * IntAsBigInt(i);
+        }
+       return x;
+    }
 
 }
 
--- a/Standard/src/Math/Functions.qs
+++ b/Standard/src/Math/Functions.qs
@@ -658,7 +658,7 @@ namespace Microsoft.Quantum.Math {
     }
     
     /// # Summary
-    /// Returns a factorial as an int.
+    /// Returns a factorial of a given number.
     ///
     /// # Description
     /// Returns the factorial as Integer, given an input of $n$ as an Integer.
@@ -729,4 +729,3 @@ namespace Microsoft.Quantum.Math {
     }
 
 }
-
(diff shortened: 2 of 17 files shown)


cgranade:
Thanks for the contribution! I'll go on and give a preliminary review. As a heads-up, since this modifies the public APIs for the Q# standard library, we'll need to review that API change through the [API review process](https://github.com/microsoft/QuantumLibraries/wiki/Q%23-API-Design-Process).

Thank you!

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
...
+    /// Returns the factorial as Integer, given an input of $n$ as an Integer.
+    /// Maximum input is |20| to return as Int. For inputs greater than 20, use FactorialB(n).
+    ///
+    /// # Input
+    /// ## $n$
+    /// An Int between -20, 20.
+    ///
+    /// # Output
+    /// The factorial of the provided input with the datatype Int.
+    function FactorialI (n : Int) : Int
```
cgranade, line 673: It can be helpful to mention in documentation comments other functions and operations that act similarly up to what types they accept as input:
```suggestion
    /// The factorial of the provided input with the datatype Int.
    ///
    /// # See Also
    /// - Microsoft.Quantum.Math.FactorialL
    function FactorialI (n : Int) : Int
```

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
             return output;
         }
     }
+    
+    /// # Summary
+    /// Returns a factorial as an int.
```
cgranade, line 661: Generally, you can leave out return types and the like from API comments, as the API documentation generation tool provides those when compiling the library.

```suggestion
    /// Returns a factorial of a given number.
```

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
...
+    /// # Description
+    /// Returns the factorial as Big Integer, given an input of $n$ as an Integer.
+    ///
+    /// # Input
+    /// ## $n$
+    /// A whole number of any size, positive or negative.
+    ///
+    /// # Output
+    /// The factorial of the provided input with the type BigInt.
+    function FactorialB (n : Int) : BigInt
```
cgranade, line 708: We've generally used the suffix "B" to indicate Bool, and "L" (for "long") to indicate BigInt:
```suggestion
    function FactorialL(n : Int) : BigInt
```

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
...
+    /// Maximum input is |20| to return as Int. For inputs greater than 20, use FactorialB(n).
+    ///
+    /// # Input
+    /// ## $n$
+    /// An Int between -20, 20.
+    ///
+    /// # Output
+    /// The factorial of the provided input with the datatype Int.
+    function FactorialI (n : Int) : Int
+    {
```
cgranade, line 674: Very minor code style thing, but we've generally put braces on the same line in Q# (next-line style is preferred by languages like C#).

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
...
+        mutable an = 1;
+        mutable x = 1;
+        if (n < 0 ){
```
cgranade, line 677: Another minor formatting nit, but parens aren't required around `if` conditions as of the February release.

```suggestion
        if n < 0 {
```

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
...
+        }
+        elif(n == 0){
+            return x;
+        }
+        elif(n >= 21){
+            fail "Largest factorial an Int can hold is |20|. Use Factorial BigInt.  \n \n >FactorialB(n) .";
+        }
+        else{
+            set an = n;
+        }
```
cgranade, line 689: Similarly here:
```suggestion
        elif n == 0 {
            return x;
        } elif n >= 21 {
            fail "Largest factorial an Int can hold is 20!. Use Factorial BigInt.  \n \n >FactorialB(n).";
        } else {
            set an = n;
        }
```

TheMagicNacho:
@cgranade 
Great! Happy to help.
Let me know if there are any issues or changes required.

Standard/src/Math/Functions.qs
```diff
@@ -656,6 +656,74 @@ namespace Microsoft.Quantum.Math {
...
+            set an = AbsI(n);
+            set x = IntAsBigInt(-1);
+        }
+        elif(n == 0){
+            return x;
+        }
+        else{
+            set an = n;
+        }
+        for i in  1 .. an {
```
cgranade, line 722: This is a very reasonable way to compute factorials for small integers, but can be very slow for larger inputs; it may be worth looking to see what techniques can improve performance for larger inputs. For instance, the GNU MP manual notes that they factor out powers of 2 from even terms appearing in the expansion of each factorial (https://web.archive.org/web/20130314174653/http://gmplib.org/manual/Factorial-Algorithm.html; link via Wikipedia). Chapter 6 of http://numerical.recipes/book/book.html also has some good suggestions for how to compute approximate factorials using double-precision floating point representations that may be useful here.

cgranade (commented):
This looks like a great start, thank you! I've taken just a preliminary pass for now, and most of my comments are fairly minor. I think the only more major that that I had was with taking factorials of larger inputs, as that can get pretty slow as the number of bits needed to represent the accumulator grows.

Happy to expand on or explain any of my comments more, if that would be helpful. Thanks again for your contribution!

Standard/src/Math/Functions.qs
TheMagicNacho: That does seem like a fun problem to solve. But do you think developers will need access to factorials larger than 20?
If not, we can drop the Long function.

cgranade: > How frequently do you think developers will need access to factorials larger than 20?

That's a fair point, yeah. I don't have data available about how often that in particular will be needed, but it's a reasonable question, I agree. My initial thought is that having the BigInt version is still reasonable, but it may be worth implementing at least the odd-factorial approach. I'm happy to do so on our end, but in case you're interested in taking that on, I think you could do it by having a second internal-only function to compute 𝑛!! = (𝑛)(𝑛 − 2)(𝑛 − 4)… for odd inputs:

```
/// # Summary
/// Given an odd integer, returns the odd-factorial
/// of that integer (also known as the double-factorial).
///
/// # Example
internal function OddFactorialL(n : Int) : BigInt {
     Fact(n % 2 == 1, "Expected n to be odd.");
     mutable acc = 1L;
     for i in 1..2..n {
          set acc *= i;
     }
     return acc;
}
```

At that point, for larger `n`, could use a recursive approach to implementing `FactorialL`. Some rough pseudocode for that (not claimed to catch edge cases and the like at all, or to have gotten all the factors of two correct):

```
function FactorialL(n : Int) : BigInt {
     if n <= 20 {
         // Use the direct approach for small n.
     } elseif n % 2 == 1 {
         return OddFactorialL(n) * FactorialL(n // 2) <<< ((n // 2) - 1);
     } else {
         return n * FactorialL(n - 1);
     }
}
```

TheMagicNacho (commented):
Changed the output from Long to Double, which outputs in scientific notation for larger numbers. And I find that more helpful in most applications.

Also, I found an approximation formula. and it was accurate within 1/ n^5 when I tested it.

Standard/src/Math/Functions.qs
TheMagicNacho: Hey @cgranade

Sorry for the slow reply.
I implemented your suggested changes.

I tested a couple different types of fast factoring algorithms, and the one I committed performed the best. Your recommendation was correct, using the even/odd method out performed the two other algorithms I tried. 

While I was working on whole numbers I also came across an approximation that works really well. This function returns a double and has a standard deviation of 1/n^5. So I implemented it as FactorialD, should a developer need that capability. This function is also nice because it outputs scientific notation. 

During testing I also found the need for BigIntAsString. So I made that and committed it to the convert namespace. 

One thing I need to change, is refactor all the braces. All of mine are on the next line down (old habit). I'll try to work on that tomorrow night. 

Let me know if there's anything else I can help with.

cgranade:
@TheMagicNacho: My apologies for the delay! In any case, I've taken a quick look, and this looks like it's really coming along, thanks for the contribution! If it's alright with you, I'd like to bring this into a new feature branch for working on #448 instead of `main`, so that I can integrate your work here with support for gamma / log-gamma and binomial coefficients. Thanks!

TheMagicNacho:
> @TheMagicNacho: My apologies for the delay! In any case, I've taken a quick look, and this looks like it's really coming along, thanks for the contribution! If it's alright with you, I'd like to bring this into a new feature branch for working on #448 instead of `main`, so that I can integrate your work here with support for gamma / log-gamma and binomial coefficients. Thanks!

@cgranade Seems fine with me. I was actually thinking gamma functions seem like the next logical step. Do you need me to re-base or can you merge it from here?

TheMagicNacho:
>  If it's alright with you, I'd like to bring this into a new feature branch for working on #448 instead of `main`

@cgranade Seems fine with me. I was actually thinking that gamma functions would be the next logical step. Do you need me to re-base to that branch, or can you merge it from here?

cgranade:
> > If it's alright with you, I'd like to bring this into a new feature branch for working on #448 instead of `main`
> 
> @cgranade Seems fine with me. I was actually thinking that gamma functions would be the next logical step. Do you need me to re-base to that branch, or can you merge it from here?

Sounds great! I can change the target branch from my side, then; will take that as a starting point for #448. Thanks for your contribution!

.gitignore
```diff
@@ -333,3 +333,7 @@ __pycache__/
 *.odx.cs
 *.xsd.cs
 
+/ionqTest
```
cgranade, line 336: We may want to remove these; looks like these entries are likely specific to your testing setup?

cgranade (approved):
Other than one minor comment, this looks good to me for bringing onto a feature branch; will add the other parts of #448 from there. Thank you!

.gitignore
TheMagicNacho: Yes. You're correct. We can delete these lines.

cgranade:
Removing API review label, since this now targets a feature branch for further development.

PR #443, opened 2021-04-12
Work around for compiler QIR generaiton issue

swernli:
This works around https://github.com/microsoft/qsharp-compiler/issues/964 by avoiding passing `Length<'T>` as a callable paramter, where `'T` is an inherited generic type.


cgranade (approved):
Looks good to me, thanks for adding this workaround! 💕

PR #446, opened 2021-04-21
Improve API documentation for ApplyUnitary

tcNickolas:
* Add usage sample
* Render the formulas in description as formulas

### DIFF
--- a/Standard/src/Synthesis/UnitaryDecomposition.qs
+++ b/Standard/src/Synthesis/UnitaryDecomposition.qs
@@ -45,16 +45,31 @@ namespace Microsoft.Quantum.Synthesis {
     }
 
     /// # Summary
-    /// Applies gate defined by 2^n x 2^n unitary matrix.
+    /// Applies gate defined by a $2^n \times 2^n$ unitary matrix.
     ///
     /// Fails if matrix is not unitary, or has wrong size. 
     ///
     /// # Input
     /// ## unitary
-    /// 2^n x 2^n unitary matrix describing the operation. 
-    /// If matrix is not unitary or not of suitable size, throws an exception.
+    /// A $2^n \times 2^n$ unitary matrix describing the operation. 
+    /// If the matrix is not unitary or not of suitable size, throws an exception.
     /// ## qubits
-    /// Qubits to which apply the operation - register of length n.
+    /// Qubits to which apply the operation - a little-endian register of length n.
+    /// 
+    /// # Example
+    /// The following operation will be equivalent to applying the Hadamard gate to the given qubit:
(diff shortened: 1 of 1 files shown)


Standard/src/Synthesis/UnitaryDecomposition.qs
```diff
@@ -45,16 +45,31 @@ namespace Microsoft.Quantum.Synthesis {
     }
 
     /// # Summary
-    /// Applies gate defined by 2^n x 2^n unitary matrix.
+    /// Applies gate defined by a $2^n \times 2^n$ unitary matrix.
```
cgranade, line 48: We've tended to avoid LaTeX notation in Summary sections, as that can confuse the VS 2019 client. I fully agree with adding in LaTeX notation at line 54, though.

Standard/src/Synthesis/UnitaryDecomposition.qs
```diff
@@ -45,16 +45,31 @@ namespace Microsoft.Quantum.Synthesis {
...
-    /// 2^n x 2^n unitary matrix describing the operation. 
-    /// If matrix is not unitary or not of suitable size, throws an exception.
+    /// A $2^n \times 2^n$ unitary matrix describing the operation. 
+    /// If the matrix is not unitary or not of suitable size, throws an exception.
     /// ## qubits
-    /// Qubits to which apply the operation - register of length n.
+    /// Qubits to which apply the operation - a little-endian register of length n.
+    /// 
+    /// # Example
+    /// The following operation will be equivalent to applying the Hadamard gate to the given qubit:
```
cgranade, line 60: This is a great example, thanks for adding it! Would it be worth noting that this way of applying an `H` operation to a given qubit is in general more expensive than calling `H` directly?

cgranade (approved):
Approving with two minor suggestions. Thanks for the improvements!

Standard/src/Synthesis/UnitaryDecomposition.qs
tcNickolas: How about Unicode, then? 2ⁿ × 2ⁿ?

cgranade: That should work, yeah. It's a bit awkward, I know, sorry...

tcNickolas: Not particularly so :-) The project versions of the Katas still relies on Unicode quite extensively, since they have to be plain text with no LaTeX support. There are some formulas that I'd really prefer not to write in Unicode, but this one is quite benign :-)

PR #466, opened 2021-06-14
Fix API docs for MultiplyAndAddByModularInteger

tcNickolas:
I wonder if we should add a usage example as well, something along the lines of "If you want to compute |(3+2⋅x)mod 5⟩, you need to allocate an extra register sum in state |3⟩ and call MultiplyAndAddByModularInteger(2, 5, x, sum)"?


cgranade (approved):
Looks good to me. I agree with adding examples, but perhaps we can defer that for now?

PR #469, opened 2021-07-13
Fix API docs for MeasureInteger

Borris-the-real-OG:
Removed reference to deprecated operation MeasureIntegerBE.


cgranade (approved):
Thank you for the fix on this, @Borris-the-real-OG!

PR #473, opened 2021-07-19
Fixes #346.

### LINKED ISSUE #346: Partitioned creates an extra zero-length partition if partitioning is exact
**Describe the bug**
If I provide [Partitioned](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arrays.partitioned) an array of length N and partition sizes of total length N, the return contains an extra partition of length 0 in the end. This makes it inconvenient to use: if I want to split an array into K parts of equal length ([example](https://github.com/microsoft/QuantumKatas/blob/ddc5d86742fffb09afc7e08d931d33c1be2c4870/GraphColoring/ReferenceImplementation.qs#L45)), I need to remember to ask for K-1 partitions. This behavior is also not documented in API docs, so I don't think it's intended.

**To Reproduce**
The following code:
```qsharp
    let p = Partitioned([2, 2], [1, 2, 3, 4]);
    Message($"{p}");
```
outputs `[[1,2],[3,4],[]]`.

**Expected behavior**
I would expect `[[1,2],[3,4]]`, without the extra empty partition.

msoeken:


Standard/src/Arrays/Arrays.qs
```diff
@@ -324,13 +324,19 @@ namespace Microsoft.Quantum.Arrays {
     /// # Output
     /// Multiple arrays where the first array is the first `nElements[0]` of `arr`
     /// and the second array are the next `nElements[1]` of `arr` etc. The last array
-    /// will contain all remaining elements.
+    /// will contain all remaining elements. If the partition is split exactly, the
```
tcNickolas, line 327:
```suggestion
    /// will contain all remaining elements. If the array is split exactly, the
```

Standard/src/Arrays/Arrays.qs
```diff
@@ -324,13 +324,19 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// last array will be the empty array, indicating there are no remaining elements.
+    /// The thought is that `Tail(Partitioned(...))` should always return the remaining
```
tcNickolas, line 329: I would make the phrasing here a bit more formal: "Most(Partitioned(...)) will always return ..., and Tail(Partitioned(...)) will always return ...". Also, there might be some extra ` symbols before Most

tcNickolas (approved):
A couple of small phrasing suggestions, other than that, looks great, thank you for taking care of this!

Standard/src/Arrays/Arrays.qs
```diff
@@ -324,13 +324,19 @@ namespace Microsoft.Quantum.Arrays {
...
+    /// elements, while ```Most(Partitioned(...))` should always return the complete
+    /// partitions of the array.
```
msoeken, line 331:
```suggestion
    /// In other words, `Tail(Partitioned(...))` will always return the remaining
    /// elements, while `Most(Partitioned(...))` will always return the complete
    /// partitions of the array.
```

PR #484, opened 2021-09-21
Fix broken links to API docs

tcNickolas:
@"" style of links doesn't seem to work if the reference is in lowercase.
Includes fixing broken formulas in PurifiedMixedState.

### DIFF
--- a/Standard/src/AmplitudeAmplification/AmplitudeAmplification.qs
+++ b/Standard/src/AmplitudeAmplification/AmplitudeAmplification.qs
@@ -149,7 +149,7 @@ namespace Microsoft.Quantum.AmplitudeAmplification {
     /// ## phases
     /// A set of phases describing the partial reflections at each step of the
     /// amplitude amplification algorithm. See
-    /// @"microsoft.quantum.amplitudeamplification.standardreflectionphases"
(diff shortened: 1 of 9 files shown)


tcNickolas:
The formulas at API page for https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.preparation.purifiedmixedstatewithdata are also broken, but I couldn't figure out what caused it at a glance.

cgranade:
> @"" style of links doesn't seem to work if the reference is in lowercase.

That's right, I don't think that ever worked properly; I had thought I fixed most of that with the new docgen tool, but thanks for catching this one as well. For the most part, `<xref:Fully.Qualified.Name>` seems to work most reliably, but `@"Fully.Qualified.Name"` should work as well.

> Includes fixing broken formulas in PurifiedMixedState.

Thanks for catching that! Wish we had a better way of previewing rendering... ah, well.

cgranade:
> The formulas at API page for https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.preparation.purifiedmixedstatewithdata are also broken, but I couldn't figure out what caused it at a glance.

I'll take a look; that used to work, but I'm guessing there was a change to how Markdig escapes things before they get to MathJax.

cgranade (approved):

PR #495, opened 2021-11-30
Fix formula in Fold documentation

tcNickolas:
Added `...` indicating multiple function applications to match `...` indicating multiple arguments.


msoeken (approved):
Thank you.

PR #516, opened 2022-02-01
Update to version 0.22.187631

msoeken:


NOTICE.txt
```diff
@@ -1,28 +1,28 @@
...
-
----------------------------------------------------------
-
-FluentAssertions 5.10.3 - Apache-2.0
-
-
-(c) 2008 VeriSign, Inc.
-Copyright Dennis Doomen
-
+NOTICES AND INFORMATION
```
tcNickolas: Should this file be part of the change? I couldn't find any difference in what GitHub highlights as diff

tcNickolas (approved):
Looks good, just one file I'm not sure about

NOTICE.txt
msoeken: I think there is an error in Github. The change is just in one line for the license of Q# compiler.

PR #517, opened 2022-02-03
Updates to new array creation syntax

tcNickolas:
This includes changes done automatically by the Q# Formatter and some changes I did by hand on top of that. The remaining warnings include arrays of qubits and arrays of generic type, fixing those requires some careful thought, since the new syntax doesn't apply to those.


msoeken (approved):
Thank you, @tcNickolas

Issue #483, opened 2021-09-16
Random, Intrinsic and Canon namespaces are missing summaries


tcNickolas:
Having a summary of the namespace makes it easier for the reader to orient themselves in its functionality.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 114ea174-7b11-668d-d1e6-dad38bd321df
* Version Independent ID: 13c4798d-91cc-47dd-c02d-1897a294d7ae
* Content: [Microsoft.Quantum.Intrinsic namespace - Q# reference - Microsoft Quantum](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic)
* Content Source: [api/qsharp/microsoft.quantum.intrinsic.md](https://github.com/MicrosoftDocs/quantum-api/blob/live/api/qsharp/microsoft.quantum.intrinsic.md)
* Product: **qsharp**
* GitHub Login: @ricardo-espinoza

tcNickolas:
Changes to both libraries and runtime are in, so the summaries should appear with the next release.

PR #519, opened 2022-02-05
Add namespace summaries and descriptions.

### LINKED ISSUE #496: Some API docs pages should point to built-in operators docs
I've seen people use functions from M.Q.Math or M.Q.Bitwise namespaces in contexts in which the regular built-in operators would work as well. I'm not sure whether it was deliberate or indicated lack of discoverability of information on the built-in operators. I think we could improve discoverability of operators by adding links to the matching pages in language documentation. For example, https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.bitwise could point to https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/expressions/bitwiseexpressions, https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.math - to https://docs.microsoft.com/en-us/azure/quantum/user-guide/language/expressions/arithmeticexpressions

tcNickolas:
This fixes #496, and includes part of the #483.

### DIFF
--- a/Standard/src/Canon/Properties/NamespaceInfo.qs
+++ b/Standard/src/Canon/Properties/NamespaceInfo.qs
@@ -1,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// # Summary
+/// This namespace contains a variety of useful operations and functions.
+///
+/// # Description
+/// To learn more about the operations in this namespace, see 
+/// [Higher-Order Control Flow](xref:microsoft.quantum.libraries.overview-standard.control-flow) and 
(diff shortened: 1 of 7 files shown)


Standard/src/Canon/Properties/NamespaceInfo.qs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
msoeken, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

msoeken (approved):
Thanks @tcNickolas

PR #525, opened 2022-02-16
Fix #511 for fixed point amplitude amplification

### LINKED ISSUE #511: An error in the function FixedPointReflectionPhases
**Describe the bug**
FixedPointReflectionPhases returns NaN phases.
https://github.com/microsoft/QuantumLibraries/blob/main/Standard/src/AmplitudeAmplification/StandardAlgorithms.qs#L53

**Expected origin of the bug**
At line 59: let beta = Cosh((1.0 / nQueriesDouble) * ArcCosh(Sqrt(successMin)));
ArcCosh is called on a value less than 1.

**Expected fix**
From https://arxiv.org/pdf/1409.3305.pdf: gamma^{-1} = T_{1/L} (1/delta).
I think that the Cosh and ArcCosh should be Cos and ArcCos instead.
(Other fixes may be needed too).

**Additional context**
Thanks for the great Amplitude Amplification library!

msoeken:
This PR will fix #511

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.g.cs
 bin/
 obj/
+docs/
 Documentation/Help/
 packages/
 *.obj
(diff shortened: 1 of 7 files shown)


msoeken:
> LGTM. Marking as comment only since it's still in draft. Only suggestion might be to add a new unit test to cover this case. Thanks!

I agree, there are still some more issues with the current implementation, so I haven't pushed my test case yet such that the pipeline does not break.

cgranade (approved):

PR #529, opened 2022-02-18
Adapting to new array syntax

msoeken:


tcNickolas (approved):
https://github.com/microsoft/qsharp-language/blob/main/Approved/2-enhanced-array-literals.md#impact-on-existing-mechanisms says that the `Default` function will also be deprecated eventually. I wonder if it makes sense to try and rewrite these cases to start with an empty array literal and append new elements to it to save ourselves another iteration down the road? At a glance, the trickiest one is `Interleaved`, the rest seem to assign elements in order.

I'll approve this PR in case you want to use `Default` for now - the change with it looks good, my only concern is that the function is going to be deprecated soon.

msoeken:
Thanks @tcNickolas. I will leave it with `Default` for now and make another pass once `Default` is deprecated.

PR #530, opened 2022-02-18
Fixes #494

### LINKED ISSUE #494: `qdk-chem convert` produces invalid broombridge format
**Describe the bug**
The command line
```
cd QuantumLibraries/Chemistry/tests/TestData/FciDump
qdk-chem convert --from fcidump --to broombridge h2o_ccpvdz.fcidump  --out h2o_cpvdz.yaml
```
produces a yaml file that does not pass validation via the [schema validator](https://github.com/microsoft/Quantum/tree/main/Chemistry/Schema).

Issues include:
 - version not of type "string"
 - missing "index_convention" key

**To Reproduce**
See above.

**Expected behavior**
The conversion tool should always produce valid broombridge yaml.

**System information**
 - OS: WSL2
 - microsoft.quantum.chemistry.tools      0.20.2110171573      qdk-chem

msoeken:


ltalirz:
Thanks a lot @msoeken !

I haven't really looked into the codebase much before, so perhaps @cgranade is better placed to comment on whether this is the right way to fix this. 

The only thing I'm wondering: would it be useful to add
```
cd QuantumLibraries/Chemistry/tests/TestData/FciDump
qdk-chem convert --from fcidump --to broombridge h2o_ccpvdz.fcidump  --out h2o_cpvdz.yaml
```
as a test, followed by schema validation?
That way we exclude such issues for the future.

Anyhow, I realize that the schema and the validator live in a different package, so perhaps that makes this difficult.

Chemistry/src/DataModel/Serialization/Broombridge/BroombridgeDataStructurev0.1.cs
```diff
@@ -362,7 +365,11 @@ public void Write(IEmitter emitter, ObjectSerializer nestedObjectSerializer)
                             entry.Item1.Select((idx) => (object)idx).Concat(new object[] { entry.Item2 })
                         )
                         .ToList()
-                });
+                };
+                if (IndexConvention != null) {
```
cgranade, line 369: [nit]
```suggestion
                if (IndexConvention != null)
                {
```

Chemistry/src/Tools/Convert.cs
```diff
@@ -72,6 +67,7 @@ public static class Convert
             // to represent data between loading and saving.
             var data = Load(reader, from);
             Save(data, writer, to);
+            writer.Close();
```
cgranade, line 70: This could have unintended consequences since the writer is passed in by the caller; if we want to have the call to `Close` here, that should be very clearly spelled out in API documentation comments to avoid confusion if a user expects the `writer` object to still be valid for further writing after this call.

cgranade (approved):
LGTM. With respect to validating the output, I agree, but there's some subtlety there we should be careful of. First and foremost, we should make sure to validate the YAML documents against the value of `$schema` rather than against the schema as stored in any particular repository, as that mirrors the workflow that a consumer would use in parsing the output of the `qdk-chem` tool.

For approaches to validating output in CI, perhaps we could consider using a .NET integration test backed by a package like https://www.nuget.org/packages/NJsonSchema? Such a test would effectively assert that output documents obey the schema that they advertise with `$schema`.

Chemistry/src/Tools/Convert.cs
msoeken: Good point, I will remove this again. I noticed that when implementing these changes the file was corrupted when specifying an existing filename and when the contents change. That's why I added a `Close` here.

PR #532, opened 2022-02-25
Unify Markdown for Examples section

tcNickolas:
Following up on discussion in #531, let's use `# Examples` consistently


msoeken (approved):
Thanks a lot, @tcNickolas

PR #541, opened 2022-03-15
Fixes #367

### LINKED ISSUE #367: AllowAtMostNCallsCA does not differentiate between controlled and not-controlled variants of a unitary
**Describe the bug**
AllowAtMostNCallsCA does not differentiate controlled and not-controlled variants of a unitary.  While, let say, X and Controlled X have similar behavior, they are still different unitary and should be considered as so using AllowAtMostNCallsCA. 

This was discovered while trying to replace Microsoft.Quantum.Katas.CounterSimulator by simple QuantumSimulator in the QuantumKatas repo

**To Reproduce**
The following code raise the AllowAtMostNCallsCA exception: 

```qsharp
within {
    AllowAtMostNCallsCA(2, Controlled X, "You are allowed to call Controlled X exactly once");
} apply {
    // Apply the operation
    op();
}

operation op() : Int {

    using ((control, eigenstate) = (Qubit(), Qubit())) {
        X(eigenstate);

        within {
            H(control);     
        } apply {
            Controlled X([control], eigenstate);
        }

        let eigenvalue = M(control) == Zero ? 1 | -1;
        Reset(eigenstate);
        return eigenvalue;
    }
}
```

The raised exception:  Unhandled exception. Microsoft.Quantum.Simulation.Core.ExecutionFailException: Operation Microsoft.Quantum.Intrinsic.X was called more than the allowed 1 times:
You are allowed to call Controlled X exactly once

**Expected behavior**
The exception should not be raised when doing call to X while AllowAtMostNCallsCA is set to Controlled X.
In other words, X and Controlled X should be considered 2 different operations.

**System information**

QDK 0.13.20102604

msoeken:


Standard/src/Diagnostics/Emulation/AllowOperationCalls.cs
```diff
@@ -65,7 +74,7 @@ public Native(IOperationFactory m) : base(m)
                     {
                         if (IsSelf(callable)) return;
                         callStack = callStack.Push(callable.FullName);
-                        if (callable.FullName == op.FullName)
```
tcNickolas, line 68: May I request a comment here? It took me a bit of thought to figure out what's happening here, so something like "Count all callables that are the restricted callable or its variants" might be a nice note to leave for our future selves

Standard/tests/Diagnostics/AllowTests.qs
```diff
@@ -64,6 +64,39 @@ namespace Microsoft.Quantum.Tests {
...
+            X(q2);
+            Controlled X([q1], q2);
+            X(q2);
+        }
+    }
+
+    @Diag.Test("QuantumSimulator")
+    operation CheckAllowNCallsDistinguishAdjoint() : Unit {
+        within {
+            Diag.AllowAtMostNCallsCA(2, Adjoint S, "Too many calls to Adjoint S.");
```
tcNickolas, line 84: I would add a restriction on S and on Controlled S to make sure to cover (and document) all variants

tcNickolas (approved):
Looks good! I left a couple of comments, mostly on documenting the function behavior, since it's a nuanced one

PR #547, opened 2022-03-28
Add gh-sync workflow

IsraelMiles:
The gh-sync GitHub Action allows you to synchronize public GitHub issues with internal ADO tracking. It can be customized to be triggered according to a repository's specific needs, see more below.
https://github.com/microsoft/gh-sync


msoeken (approved):
Thanks!

PR #548, opened 2022-03-30
Updating projects and other components to build 0.24.201332

ricardo-espinoza:


anjbur (approved):
Looks good to me, except that the NOTICE.txt file at the root should also be updated

Issue #442, opened 2021-04-08
New functions for singly controlled operations


cgranade:
# Functions for singly controlled operations

## Conceptual overview

The Q# language provides the `Controlled` functor to control operations on quantum registers of arbitrary size. It is often conveinent, however, for operations to accept exactly one control qubit. For example, `Controlled X` has type `(Qubit[], Qubit) => Unit is Adj + Ctl`, allowing for an arbitrary number of control qubits, but the shorthand `CNOT` has type `(Qubit, Qubit) => Unit is Adj + Ctl`. This shorthand is especially useful together with combinator operations such as `ApplyToEachCA`:

```qsharp
ApplyToEachCA(CNOT, Zipped(controls, targets));
```

### Current status

For operations other than `X`, no such shorthand may exist in general. For example, on a recent stream, @crazy4pi314 used the following snippet combining `ApplyToEachCA` with `IncrementByInteger` to count the Hamming weight of a control register:

```qsharp
ApplyToEachCA(
    (Controlled IncrementByInteger)(_, (1, target)),
    Mapped(ConstantArray(1, _), controls), 
);
```

In this example, mapping `ConstantArray(1, _)` is needed to add another level of nesting to `controls`, and is only needed because 
`(Controlled IncrementByInteger)(_, (1, target))` has type `Qubit[] => Unit is Adj + Ctl` (that is, takes an arbitrary number of control qubits).

### User feedback

See above example from @crazy4pi314's stream at https://www.twitch.tv/crazy4pi314.

### New and modified functions, operations, and UDTs

```qsharp
namespace Microsoft.Quantum.Canon {

    /// # Summary
    /// Given a controllable operation, returns a controlled version of that operation
    /// accepting exactly one control qubit.
    ///
    /// # Input
    /// ## op
    /// The operation to be controlled.
    ///
    /// # Output
    /// A controlled variant of `op` accepting exactly one control qubit.
    ///
    /// # Example
    /// To add the weight (number of "1" bits) of a control register to
    /// a target register:
    /// ```qsharp
    /// ApplyToEachCA(
    ///     (SinglyControlled(IncrementByInteger))(_, (1, target)),
    ///     controls)
    /// );
    /// ```
    ///
    /// # See Also
    /// - Microsoft.Quantum.Canon.SinglyControlledA
    function SinglyControlled<'T>(op : ('T => Unit is Ctl)) : ((Qubit, 'T) => Unit is Ctl ){
        // ..
    }

[...]

PR #552, opened 2022-04-06
Implements #442

msoeken:


Standard/src/Canon/Combinators/SinglyControlled.qs
```diff
@@ -0,0 +1,68 @@
...
+    /// ApplyToEachCA(
+    ///     SinglyControlled(IncrementByInteger)(_, (1, target)),
+    ///     controls)
+    /// );
+    /// ```
+    ///
+    /// # See Also
+    /// - Microsoft.Quantum.Canon.SinglyControlledA
+    function SinglyControlled<'T>(op : 'T => Unit is Ctl) : (Qubit, 'T) => Unit is Ctl {
+        return ApplySinglyControlled(op, _, _);
```
cgranade, line 30: You may be able to write this as a lambda now:

```suggestion
        return (ctrl, originalInput) => Controlled op([ctrl], originalInput);
```

Not sure if that generates the right characteristics, as I've not tested that yet, sorry.

Standard/tests/Canon/Combinators/SinglyControlled.qs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Tests {
+    open Microsoft.Quantum.Arithmetic;
+    open Microsoft.Quantum.Canon;
+    open Microsoft.Quantum.Diagnostics;
+    open Microsoft.Quantum.Intrinsic;
+
+    @Test("ToffoliSimulator")
+    operation TestSinglyControlled() : Unit {
```
cgranade, line 11: For testing, this is great, but I may also suggest an AssertOperationsEqualReferenced call for something like `H`:

```qsharp
AssertOperationsEqualReferenced(2,
     (qs) => SinglyControlled(H)(qs[0], qs[1]),
     (qs) => Controlled H([qs[0]], qs[1])
);
```

cgranade (approved):
Approved with minor comments. Thanks!

Standard/src/Canon/Combinators/SinglyControlled.qs
msoeken: That's awesome! And the recent commit confirms it works.

PR #553, opened 2022-04-07
Fix #546

### LINKED ISSUE #546: Refactor `Mapped` to avoid use of `Default`
**Describe the bug**
`Default` uses deprecated new array syntax to create default and often invalid values for generic types. Because `Mapped` uses `Default` to pre-allocate the array being mapped into, it can inadvertently trigger asserts in runtimes that check for valid content on creation of values. We should find a way to refactor `Mapped` to protect against this.

One possible suggestion for a new implementation (courtesy of @cgranade):
```qsharp
// pseudocode
function Mapped<'T, 'O>(fn : 'T -> 'O, arr : 'T[]) : 'O {
    if Length(arr) == 0 {
        return [];
    }
    let first = fn(arr[0]);
    mutable retval = [first; size=Length(arr)];
    for idx in 1..Length(arr) - 1{
        set retval w/ idx <- fn(arr[idx]);
    }
    return retval;
}
```
Related:
https://github.com/microsoft/qsharp-runtime/issues/977
https://github.com/microsoft/qsharp-runtime/issues/975

msoeken:
Fixes #546 and removes all other occurrences of `Default`; also removes two warnings.

### DIFF
--- a/MachineLearning/src/Training.qs
+++ b/MachineLearning/src/Training.qs
@@ -87,8 +87,7 @@ namespace Microsoft.Quantum.MachineLearning {
         trainingSchedule : SamplingSchedule,
         validationSchedule : SamplingSchedule
     ) : (SequentialModel, Int) {
-        mutable bestSoFar = Default<SequentialModel>()
-                            w/ Structure <- (Head(models))::Structure;
+        mutable bestSoFar = SequentialModel((Head(models))::Structure, [], 0.0);
         mutable bestValidation = Length(samples) + 1;
 
         let features = Mapped(_Features, samples);
--- a/MachineLearning/tests/StructureTests.qs
+++ b/MachineLearning/tests/StructureTests.qs
@@ -10,33 +10,20 @@ namespace Microsoft.Quantum.MachineLearning.Tests {
 
     @Test("QuantumSimulator")
     function NQubitsRequiredFact() : Unit {
-        let model = Default<ML.SequentialModel>()
-            w/ Structure <- [
-                ML.ControlledRotation((3, [7, 9]), PauliX, 0),
-                ML.ControlledRotation((8, []), PauliY, 1)
-            ];
+        let model = ML.SequentialModel([
+            ML.ControlledRotation((3, [7, 9]), PauliX, 0),
+            ML.ControlledRotation((8, []), PauliY, 1)
+        ], [], 0.);
         let actual = ML.NQubitsRequired(model);
         EqualityFactI(actual, 10, "Wrong output from NQubitsRequired.");
     }
 
     function ExampleModel() : ML.SequentialModel {
-        return Default<ML.SequentialModel>()
-            w/ Structure <- [
-                Default<ML.ControlledRotation>()
-                    w/ TargetIndex <- 2
-                    w/ ControlIndices <- [0]
-                    w/ Axis <- PauliX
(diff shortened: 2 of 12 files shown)


MachineLearning/src/Training.qs
```diff
@@ -87,8 +87,7 @@ namespace Microsoft.Quantum.MachineLearning {
         trainingSchedule : SamplingSchedule,
         validationSchedule : SamplingSchedule
     ) : (SequentialModel, Int) {
-        mutable bestSoFar = Default<SequentialModel>()
```
cgranade, line 90: This may be worth an extended discussion. Before deprecating `Default`, it was somewhat of an idiom to use `Default<UdtType>() w/ ItemName <- foo` to kind of get named-item initializers instead of relying purely on positional construction. It may be good to have a language feature similar to Rust's `{}`-style structs as an alternative way of creating UDT values.

MachineLearning/tests/StructureTests.qs
```diff
@@ -10,33 +10,20 @@ namespace Microsoft.Quantum.MachineLearning.Tests {
...
+            ML.ControlledRotation((3, [7, 9]), PauliX, 0),
+            ML.ControlledRotation((8, []), PauliY, 1)
+        ], [], 0.);
         let actual = ML.NQubitsRequired(model);
         EqualityFactI(actual, 10, "Wrong output from NQubitsRequired.");
     }
 
     function ExampleModel() : ML.SequentialModel {
-        return Default<ML.SequentialModel>()
-            w/ Structure <- [
```
cgranade, line 24: Similarly here, I think the original code was more readable even if less correct... from that perspective, I agree with your change but would like to kick off a discussion on more readable alternatives going forward?

Standard/src/Arrays/Reductions.qs
```diff
@@ -43,7 +43,8 @@ namespace Microsoft.Quantum.Arrays {
     /// ```
     function CumulativeFolded<'State, 'T>(fn : (('State, 'T) -> 'State), state : 'State, array : 'T[]) : 'State[] {
         mutable current = state;
-        mutable result = [Default<'State>(), size = Length(array)];
+        // initialize with current, and then overwrite in loop
+        mutable result = [current, size = Length(array)];
```
cgranade, line 47: Does this need an early exit if `array` is empty, or is that implied by `size = 0`?

cgranade (approved):

MachineLearning/tests/StructureTests.qs
msoeken: I agree that the original code was more readable. Do you suggest to keep the version using `Default`?

Standard/src/Arrays/Reductions.qs
msoeken: That is implied by `size = 0` since we do not need to access or compute some element based on the `0` index in this case.

cgranade: Makes sense, wasn't entirely sure from a readthrough so wanted to double-check. Thanks!

MachineLearning/tests/StructureTests.qs
cgranade: I think getting rid of `Default` is worth temporarily having less readable code, so no disagreement with your PR so much as that I want to keep momentum on a longer-term solution, I guess?

Issue #555, opened 2022-04-09
Add SubtractFxP and InvertFxP to the Numerics package


msoeken:
# Add SubtractFxP and InvertFxP to the Numerics package

## Conceptual overview

Add two operations to extend the fixed-point arithmetic capabilities in the Microsoft.Quantum.Numerics package.

### Current status

Such operations do not exists.

### New and modified functions, operations, and UDTs

```qsharp
namespace Microsoft.Quantum.Arithmetic {
    /// # Summary
    /// Computes `minuend - subtrahend` and stores the difference in `minuend`.
    ///
    /// # Input
    /// ## minuend
    /// The minuend of the subtraction
    /// ## subtrahend
    /// The subtrahend of the subtraction
    ///
    /// # See also
    /// - Microsoft.Quantum.Arithmetic.AddFxP
    /// - Microsoft.Quantum.Arithmetic.InvertFxP
    operation SubtractFxP(minuend : FixedPoint, subtrahend : FixedPoint) : Unit is Adj + Ctl {}

    /// # Summary
    /// Computes the additive inverse of `fp`.
    ///
    /// # Input
    /// ## fp
    /// Fixed-point number to invert
    ///
    /// # Remarks
    /// Numerical inaccuracies may occur depending on the
    /// bit-precision of the fixed-point number.
    ///
    /// # See also
    /// - Microsoft.Quantum.Arithmetic.SubtractFxP
    operation InvertFxP(fp : FixedPoint) : Unit is Adj + Ctl {}
}
```

PR #557, opened 2022-04-09
Implements #555

msoeken:
Changing target branch to feature branch, such that feature branch can be merged after API review.


Numerics/src/FixedPoint/Addition.qs
```diff
@@ -47,4 +47,46 @@ namespace Microsoft.Quantum.Arithmetic {
 
         AddI(LittleEndian(xs), LittleEndian(ys));
     }
+
+    /// # Summary
+    /// Computes the additive inverse of `fp`.
+    ///
+    /// # Input
+    /// ## fp
+    /// Fixed-point number to invert
```
tcNickolas, line 56: Looking at other API docs, we typically use full stop in the end of argument description
```suggestion
    /// Fixed-point number to invert.
```

Numerics/src/FixedPoint/Addition.qs
```diff
@@ -47,4 +47,46 @@ namespace Microsoft.Quantum.Arithmetic {
...
+        let (_, reg) = fp!;
+        Invert2sSI(SignedLittleEndian(LittleEndian(reg)));
+    }
+
+    /// # Summary
+    /// Computes `minuend - subtrahend` and stores the difference in `minuend`.
+    ///
+    /// # Input
+    /// ## subtrahend
+    /// The subtrahend of the subtraction
```
tcNickolas, line 74: I might expand on this definition - I had to look up both "subtrahend" and "minuend" in the dictionary, so we might get other users not familiar with the terms:
```suggestion
    /// The subtrahend of the subtraction - the number to be subtracted.
```

Numerics/src/FixedPoint/Addition.qs
```diff
@@ -47,4 +47,46 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    /// ## minuend
+    /// The minuend of the subtraction
```
tcNickolas, line 76:
```suggestion
    /// The minuend of the subtraction - the number from which the other is subtracted.
```

Numerics/src/FixedPoint/Addition.qs
```diff
@@ -47,4 +47,46 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    ///
+    /// # Remarks
+    /// Computes the difference by inverting `subtrahend` before and after adding
+    /// it to `minuend`.
+    ///
+    /// # See also
+    /// - Microsoft.Quantum.Arithmetic.AddFxP
+    /// - Microsoft.Quantum.Arithmetic.InvertFxP
+    operation SubtractFxP(subtrahend : FixedPoint, minuend : FixedPoint) : Unit is Adj + Ctl {
```
tcNickolas, line 85: I think this will be part of the API review, but I expect subtraction to subtract the second argument from the first one, rather than the other way around. Our functions in M.Q.Math follow the same order, MinusI(a, b) = a - b rather than the other way around.

tcNickolas (approved):
The logic looks good, I left some comments on API docs and the API itself

As a side note, we should probably revisit our Math API docs - MinusI entry describes both parameters as "The input to be subtracted." :-)

Numerics/src/FixedPoint/Addition.qs
msoeken: That's a good point for the API review. The problem is, that the minuend is the argument that is updated. And we aim for having *target*-like quantum registers last. So here are two conflicting guidelines and one needs to pick one.

tcNickolas: For [DivideI](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arithmetic.dividei), for example, we have arguments x, y, and result, divide x by y (intuitive) and update x and result. I think counterintuitive order of arguments would be a lot worse here than updating the first argument. We do have a -= b in the classical part of the language, which also updates the first argument

msoeken: That's a good point. However, for addition we also update the second argument. But I also tend to change the order of arguments here.

msoeken: I changed order both in here and in the API review issue.

Issue #559, opened 2022-04-11
FixedPoint conversion functions


msoeken:
# FixedPoint conversion functions

## Conceptual overview

This adds two functions to convert fixed-point representations to and from `Bool[]`. This extracts the underlying logic for `PrepareFxP` and `MeasureFxP` as functions.

### Current status

This functions are only implicitly available and would require qubit allocation in order to access them.

### New and modified functions, operations, and UDTs

```qsharp
namespace Microsoft.Quantum.Convert
{

/// # Summary
/// Computes fixed-point approximation for a double and returns it as `Bool` array.
///
/// # Input
/// ## integerBits
/// Assumed number of integerBits (including the sign big)
/// ## fractionalBits
/// Assumed number of fractionalBits
/// ## value
/// Value to be approximated
function FixedPointAsBoolArray(integerBits : Int, fractionalBits : Int, value : Double) : Bool[] {}

/// # Summary
/// Returns the double value of a fixed-point approximation from of a `Bool` array.
///
/// # Input
/// ## integerBits
/// Assumed number of integerBits (including the sign big)
/// ## bits
/// Bit-string representation of approximated number
function BoolArrayAsFixedPoint(integerBits : Int, bits : Bool[]) : Double {}

/// # Summary
/// Discretizes a double value as a double and returns its approximated representation as a double.
///
/// # Input
/// ## integerBits
/// Assumed number of integerBits (including the sign big)
/// ## fractionalBits
/// Assumed number of fractionalBits
/// ## value
/// Value to be approximated
function DoubleAsFixedPoint(integerBits : Int, fractionalBits : Int, value : Double) : Double {
    return BoolArrayAsFixedPoint(integerBits, fractionalBits, FixedPointAsBoolArray(integerBits, fractionalBits, value));
}

}
```

### Current status

TODO.

### Using proposed changes

TODO.

## Open design questions and considerations

These functions should be used in the implementation of `PrepareFxP` and `MeasureFxP`.

tcNickolas:
I missed this when I was reviewing #575, but these functions will need examples in the API docs. They're classical so examples should be simple, giving usage example and expected results for them.

PR #562, opened 2022-04-12
Fixes #560

### LINKED ISSUE #560: MeasureFxP fails to return smallest representable number
**Describe the bug**
When measuring a quantum register in a state that corresponds to the smallest representable fixed-point number, `MeasureFxP` returns `-0.0` instead of the correct value.

**To Reproduce**

Failing test case:

```qsharp
@Test("ToffoliSimulator")
operation FailingTest() : Unit {
    use qs = Qubit[4];
    let fxp = FixedPoint(2, qs);

    // bit pattern 00.00
    ApplyPauliFromBitString(PauliX, true, [false, false, false, false], qs);
    NearEqualityFactD(MeasureFxP(fxp), 0.0);

    // bit pattern 10.00
    ApplyPauliFromBitString(PauliX, true, [false, false, false, true], qs);
    NearEqualityFactD(MeasureFxP(fxp), -2.0);
}
```

**Expected behavior**
Test case should pass.

msoeken:
This also implements one function proposed in #559, however, kept `internal` as it has not been approved by API review.

### DIFF
--- a/Numerics/src/FixedPoint/Measurement.qs
+++ b/Numerics/src/FixedPoint/Measurement.qs
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 
 namespace Microsoft.Quantum.Arithmetic {
+    open Microsoft.Quantum.Arrays;
+    open Microsoft.Quantum.Canon;
     open Microsoft.Quantum.Convert;
     open Microsoft.Quantum.Measurement;
-    open Microsoft.Quantum.Math;
 
     /// # Summary
     /// Measure a fixed-point number, returns its value as Double, and resets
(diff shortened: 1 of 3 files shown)


Numerics/tests/FixedPointTests.qs
```diff
@@ -21,6 +22,33 @@ namespace Microsoft.Quantum.Numerics.ToffoliTests {
         }
     }
 
+    internal operation PrepareAsSignedAndMeasure(value : Int, fxp : FixedPoint) : Double {
+        ApplyXorInPlace(value, LittleEndian(Snd(fxp!)));
+        return MeasureFxP(fxp);
+    }
+
+    operation MeasureFxPTest() : Unit {
```
cgranade, line 30: Does this need a test attribute?
```suggestion
    @Test("QuantumSimulator")
    operation TestMeasureFxP() : Unit {
```

Numerics/src/FixedPoint/Measurement.qs
```diff
@@ -15,18 +16,8 @@ namespace Microsoft.Quantum.Arithmetic {
...
-            mutable currentRes = MResetZ(xs[i]) == (sign ? Zero | One);
-            if keepAdding {
-                set keepAdding = currentRes;
-                set currentRes = not currentRes;
-            }
-            set fpAsDouble = fpAsDouble * 0.5 + (currentRes == true ? 1. | 0.);
-        }
-        return (sign ? -1.0 | 1.0) * fpAsDouble * PowD(2.0, IntAsDouble(p-2));
+
+        let bits = Mapped(IsResultOne, ForEach(MResetZ, xs));
```
cgranade, line 20: This works great as is, but may be easier to read using new lambda syntax — could also be harder, not sure what your preference would be here!
```suggestion
        let bits = ForEach(q => MResetZ(q) == One, qs);
```

cgranade (approved):

Numerics/tests/FixedPointTests.qs
msoeken: Tests in the Numeric package are still executed via a C# test driver. I want to convert all tests to use the test attribute and remove the C# test driver in a separate PR.

Issue #465, opened 2021-06-14
Why does MultiplyI require inputs of equal length?


tcNickolas:
The implementation doesn't seem to require inputs x and y of equal length, and it seems to me that it would be convenient to be able to multiply two numbers of length N and M to get a result of length N+M.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 32fff831-2b30-3a06-bfc4-94a803f6024f
* Version Independent ID: 313b4ce3-5000-27e1-6a1f-19ecd8481bd5
* Content: [MultiplyI operation - Q# reference - Microsoft Quantum](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.arithmetic.multiplyi)
* Content Source: [api/qsharp/microsoft.quantum.arithmetic.multiplyi.md](https://github.com/MicrosoftDocs/quantum-api/blob/live/api/qsharp/microsoft.quantum.arithmetic.multiplyi.md)
* Product: **qsharp**
* GitHub Login: @ricardo-espinoza

msoeken:
Closed via #568.

PR #568, opened 2022-04-17
Implements #465

msoeken:
This makes `MultiplyI` less restrictive on the input arguments and also improves the implementation by making it less costly. It uses a new operation `ApplyAndLadder`, which is still `internal` but should go through API review. Merging this PR is not blocked on that.


Numerics/tests/IntegerHighLevelTests.qs
```diff
@@ -317,4 +311,18 @@ namespace Microsoft.Quantum.Numerics.ToffoliTests {
             }
         }
     }
-}
\ No newline at end of file
+
+    operation ExhaustiveTestHelper2NonRegularArgs(TestFunction: (Int, Int, Int, Int, Int) => Unit) : Unit {
```
tcNickolas, line 315: It is not immediately clear from the code what's the difference between regular args and non-regular args; can you add a one-line comment before each helper clarifying in which cases it should be used?

Numerics/src/Integer/Multiplication.qs
```diff
@@ -12,82 +13,142 @@ namespace Microsoft.Quantum.Arithmetic {
...
+    }
+
     /// # Summary
     /// Computes the square of the integer `xs` into `result`,
     /// which must be zero initially.
     ///
     /// # Input
     /// ## xs
-    /// $n$-bit number to square (LittleEndian)
+    /// 𝑛-bit number to square (`LittleEndian`)
```
tcNickolas, line 101: I don't think you need to remind the type of the arguments here, API docs generated have both variable name and type right above this comment, plus it includes the link to the data type so carries more info:
image
(And the same comment for the rest of the operation signatures in the PR)

tcNickolas (approved):
Looks good, just a couple of small comments to improve the look of API docs and the test code readability

PR #569, opened 2022-04-20
Refactor and remove duplicate internal operation

msoeken:
A recent PR has implemented an operation that already existed.


tcNickolas (approved):
Good point! I _thought_ I've seen a library op with "Ladder" in the title when I reviewed that PR, but I didn't check for it in the API docs - my bad!

PR #586, opened 2022-05-13
Creating global.json

anpaz:
Changes on .Net 6.0.300 requires that a local global.json includes the version of the Quantum.Sdk otherwise projects that don't have it specified fail compilation, even if there is another global.json in their parent path.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,0 +1,9 @@
+{
+    "sdk": {
+        "version": "6.0.100",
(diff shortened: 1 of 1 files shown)


msoeken:
@anpaz, could you please add some description on the reason for this PR.

ricardo-espinoza (approved):

PR #591, opened 2022-05-25
Updated to the latest QDK.

kuzminrobin:


ricardo-espinoza:
Thanks for making this update. It seems we didn't do this in April because we're going from the March release version (0.24.201332) directly to May.  LGTM.

ricardo-espinoza (approved):

IsraelMiles (approved):

Issue #594, opened 2022-06-08
Functions for smallest and largest representable fixed point


msoeken:
# Functions for smallest and largest representable fixed point

## Conceptual overview

This provides two functions to return the smallest and largest representable fixed point based on number of integer and fractional bits.

This should be added to the `Numerics` package.

### New and modified functions, operations, and UDTs

```qsharp
namespace Microsoft.Quantum.Math {

/// # Summary
/// Returns the smallest representable number for specific fixed point dimensions
///
/// # Input
/// ## integerBits
/// Number of integer bits
/// ## fractionalBits
/// Number of fractional bits
///
/// # Remark
/// The value can be computed as $-2^{p-1}$, where $p$ is the number of integer bits.
function SmallestFixedPoint(integerBits : Int, fractionalBits : Int) : Double { ... }

/// # Summary
/// Returns the largest representable number for specific fixed point dimensions
///
/// # Input
/// ## integerBits
/// Number of integer bits
/// ## fractionalBits
/// Number of fractional bits
///
/// # Remark
/// The value can be computed as $2^{p-1} - 2^{-q}$, where $p$
/// is the number of integer bits and $q$ is the number of fractional bits.
function LargestFixedPoint(integerBits : Int, fractionalBits : Int) : Double { ... }

}
```

## Open questions

Which namespace fits well?

* `Microsoft.Quantum.Arithmetic`
* `Microsoft.Quantum.Canon`
* `Microsoft.Quantum.Math`

tcNickolas:
I would add formulas to API documentation clarifying what these values are - I think there is a nice enough formula for both.
Other than that, looks good!

msoeken:
Updated the issue.

PR #606, opened 2022-07-13
Add functions for smallest and largest representable fixed point

msoeken:
Implements #594


tcNickolas (approved):
I left a couple of small suggestions for API docs formatting, other than that, looks good!

PR #617, opened 2022-08-30
Revise ApplyFunctionWithLookupTable

msoeken:

PR #618, opened 2022-08-30
Revise ApplyFunctionWithLookupTable (#617)

DmitryVasilevsky:
Taking the fix we need for release.


ricardo-espinoza (changes requested):
In general, we try to avoid doing PRs directly from main into the release branch. The correct process here would be to do something like:
- Create a new branch from the head of `release/v0.25.2208`
- Cherry-pick the change in main (3d0aea9) to that branch.
- Pull request from that branch into `release/v0.25.2208`

DmitryVasilevsky:
There were no other commits to main. However, I'll close this request and create a new one with the cherry-pick process.

PR #644, opened 2022-11-13
Remove NumSharp dependency

msoeken:
This refactors some code that deals with unitary matrix to drop the NumSharp dependency.


Standard/tests/Diagnostics/DumpTests.cs
```diff
@@ -54,6 +51,8 @@ public void DumpSReturnsCorrectMatrix()
             Assert.Equal(0.0, (double)unitary.Data[0, 1, 1], precision: 6);
             Assert.Equal(0.0, (double)unitary.Data[1, 0, 1], precision: 6);
             Assert.Equal(1.0, (double)unitary.Data[1, 1, 1], precision: 6);
+
+            Assert.Equal("Real:\n[[1.0000000000000002, 0], \n[0, 0]]\nImag:\n[[0, 0], \n[0, 1.0000000000000002]]", unitary.ToString());
```
billti, line 55: Why are the array dimensions different between implementations?

msoeken: The previous implementation uses a 3 dimensional matrix of floating-point values to represent a two-dimensional matrix of complex numbers (the extra dimension is used to represent the real and imaginary parts of the values).

The new implementation uses a two dimensional matrix of complex values (internally it uses a one-dimensional list to store everything).

billti (approved):

PR #654, opened 2023-01-14
Fix an array operation error of IdenticalFormatFactFxP and IdenticalPointPosFactFxP

weucode:
Two APIs in Fact.qs are out of work when the length of FixedPoint array is 2. Rest is proper to replace Most, in this way, this case will also be included to check.


msoeken (approved):

msoeken:
Great catch, thanks for the PR @weucode

PR #673, opened 2023-05-01
Add LocalUnivariateMinimum function parameter range determination

### LINKED ISSUE #672: LocalUnivariateMinimum parameter range positive and negative problem
# Positive and negative problems of the tolerance parameter in LocalUnivariateMinimum

## Conceptual overview

1. When calling the LocalUnivariateMinimum function, the search will never stop if the tolerance input is negative, because the bool value of ` Width(interval) > tolerance ` is always `true`.
2. When the (item1,item2) in argument `bounds`  is `item1>item2`, the bool value of `Width(interval) > tolerance` is always `false`, the program will not enter the search logic.

The above two problems are that different parameters lead to the same abnormal behavior

### Current status

1. `mutable tolerance = -1.0;`The program will never finish executing

2. `let bounds = (1.0, 0.1);` The program will never search

### New and modified functions, operations, and UDTs

1. [LocalUnivariateMinimum source code](https://github.com/microsoft/QuantumLibraries/blob/2214d89925e732bd3a10c642885699834cb975fa/Standard/src/Optimization/Univariate.qs#L73)https://github.com/microsoft/QuantumLibraries/blob/2214d89925e732bd3a10c642885699834cb975fa/Standard/src/Optimization/Univariate.qs#L73  Adding fact statement:

   `Fact(tolerance > 0.0, "The tolerance value must be positive.");`

2. [LocalUnivariateMinimum Width source code](https://github.com/microsoft/QuantumLibraries/blob/2214d89925e732bd3a10c642885699834cb975fa/Standard/src/Optimization/Univariate.qs#L28)https://github.com/microsoft/QuantumLibraries/blob/2214d89925e732bd3a10c642885699834cb975fa/Standard/src/Optimization/Univariate.qs#L28 Adding fact statement:

   `Fact(left <= right, "Left endpoint of bounds must be less than or equal to right endpoint.");`

I hope I will be allowed to submit a pr as soon as possible because this logic problem is serious.

#  Code

```C#
namespace Test {
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Math;
    open Microsoft.Quantum.Optimization;


    @EntryPoint()
    operation main () : Unit {
        //mutable tolerance = -1.0;
        mutable tolerance = 0.1;
        mutable fn = AbsD;
        let bounds = (1.0, 0.1);
        //let bounds = (0.1,1.0);
        let result = LocalUnivariateMinimum(fn, bounds, tolerance);
        Message($"{result}");
    }
}
```

AidPaike:
fix #672 

Please review issue #672 and review the pr

### DIFF
--- a/Standard/src/Optimization/Univariate.qs
+++ b/Standard/src/Optimization/Univariate.qs
@@ -5,7 +5,7 @@ namespace Microsoft.Quantum.Optimization {
     open Microsoft.Quantum.Canon;
     open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Math;
-
+    open Microsoft.Quantum.Diagnostics;
     /// # Summary
     /// Represents the result of optimizing a univariate function.
     ///
--- a/Standard/src/Optimization/Univariate.qs
+++ b/Standard/src/Optimization/Univariate.qs
@@ -25,6 +25,7 @@ namespace Microsoft.Quantum.Optimization {
     /// # Summary
     /// Returns the width of an interval.
     internal function Width(left : Double, right : Double) : Double {
+        Fact(left <= right, "Left endpoint of bounds must be less than or equal to right endpoint.");
         return right - left;
     }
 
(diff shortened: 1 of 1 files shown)


AidPaike (commented):
[Merge branch 'main' into fix-672](https://github.com/AidPaike/QuantumLibraries/commit/5b9ed015f8b8e19fd451af854e21e7c1053c2e39)

